PR #4, opened 2017-03-02
adding tests for iris tools

abit2:
adding tests for iris tools and corrected the convert_photons_to_DN and DN_to_photons function in IRISRaster class

### DIFF
--- a/irispy/iris_tools.py
+++ b/irispy/iris_tools.py
@@ -19,13 +19,12 @@ from sunpy.time import parse_time
 
 # Define some properties of IRIS detectors.  Source: IRIS instrument
 # paper.
-DETECTOR_GAIN = {"NUV": 18., "FUV": 6., "SJI": 18.}
-DETECTOR_YIELD = {"NUV": 1., "FUV": 1.5, "SJI": 1.}
+DETECTOR_GAIN = {"NUV": 18., "FUV1": 6., "FUV2": 6., "SJI": 18.}
+DETECTOR_YIELD = {"NUV": 1., "FUV1": 1.5, "FUV2": 1.5, "SJI": 1.}
 READOUT_NOISE = {"NUV": {"value": 1.2, "unit": "DN"}, "FUV": {"value": 3.1, "unit": "DN"},
                  "SJI": {"value": 1.2, "unit": "DN"}}
 
-
-def convert_DN_to_photons(data, detector_type):
+def convert_DN_to_photons(data, detector_type='NUV'):
     """Converts IRIS data number to photon counts depending on which CCD is being used.
 
     Parameters
@@ -35,6 +34,7 @@ def convert_DN_to_photons(data, detector_type):
     detector_type: `str`
         Detector/CCD upon which data was recorded.
         Can take values 'NUV', 'FUV' or 'SJI'.
+    default_detector_type : `NUV`
 
     Returns
     -------
@@ -42,10 +42,12 @@ def convert_DN_to_photons(data, detector_type):
         IRIS data in units of photon counts.
 
     """
+    # if(data.isnan):
+    #     print("HELLO")
     return DETECTOR_GAIN[detector_type]/DETECTOR_YIELD[detector_type]*data
 
 
-def convert_photons_to_DN(data, detector_type):
+def convert_photons_to_DN(data, detector_type='NUV'):
     """Converts photon counts to IRIS data number depending on which CCD is being used.
 
     Parameters
@@ -55,6 +57,7 @@ def convert_photons_to_DN(data, detector_type):
     detector_type: `str`
         Detector/CCD upon which data was recorded.
         Can take values 'NUV', 'FUV' or 'SJI'.
+    default_detector_type : `NUV`
 
     Returns
     -------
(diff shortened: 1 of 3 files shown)


abit2:
ping @DanRyanIrish

DanRyanIrish:
Thanks so much @abit2 for this PR and contributing to irispy.  I'm going through it now.

irispy/iris_tools.py
```diff
@@ -19,13 +19,12 @@
 
 # Define some properties of IRIS detectors.  Source: IRIS instrument
 # paper.
-DETECTOR_GAIN = {"NUV": 18., "FUV": 6., "SJI": 18.}
-DETECTOR_YIELD = {"NUV": 1., "FUV": 1.5, "SJI": 1.}
+DETECTOR_GAIN = {"NUV": 18., "FUV1": 6., "FUV2": 6., "SJI": 18.}
+DETECTOR_YIELD = {"NUV": 1., "FUV1": 1.5, "FUV2": 1.5, "SJI": 1.}
```
DanRyanIrish: The reason there is only one an FUV key in this dictionary, rather than an FUV1 and FUV2 is that it reflects the metadata supplied by the IRIS instrument team and other aspects of the metadata may only be labelled FUV.  Not everything is labelled FUV 1 or FUV2.  So I think this should stay FUV for the time being.  When determining what detector type to use, when can do something like ```if "FUV" in detector_type```.  This will give us the right mapping whether detector_type is "FUV", "FUV1" or "FUV2".

irispy/iris_tools.py
```diff
@@ -35,17 +34,20 @@ def convert_DN_to_photons(data, detector_type):
...
+    default_detector_type : `NUV`
 
     Returns
     -------
     data_photons: array-like
         IRIS data in units of photon counts.
 
     """
+    # if(data.isnan):
+    #     print("HELLO")
```
DanRyanIrish: Delete these lines

irispy/tests/test_iris_tools.py
```diff
@@ -1,2 +1,99 @@
 # -*- coding: utf-8 -*-
-# """Tests for functions in iris_tools.py"""
\ No newline at end of file
+# """Tests for functions in iris_tools.py"""
```
DanRyanIrish: Delete # from start of this line.  That was my fault for putting it in the first place.

irispy/tests/test_iris_tools.py
```diff
@@ -1,2 +1,99 @@
...
+source_data = np.array([[ 0.563,  1.132, -1.343],
+						[-0.719,  1.441, 1.566]])
+
+source_data1 = np.array([[1, 2, 3],
+						[4, 5, 6]])
+
+def test_convert_DN_to_photons_NUV():
+	"""
+	"""
+	# source_DN = [[ 0.563004,  1.132289, -1.343129], [-0.719726,  1.441411,  1.566724]]
```
DanRyanIrish: What's the commented out line?  Can it be deleted?  The same question for the other commented out ```source_xxx = []``` lines.

DanRyanIrish:
Thanks @abit2.  I had a quick look at your PR it looks good.  I'll let you know when I get a chance to make some more comments.

irispy/iris_tools.py
```diff
@@ -85,13 +90,17 @@ def calculate_intensity_fractional_uncertainty(data, data_unit, detector_type):
         Same shape as data.
 
     """
+    if 'FUV' in detector_type:
+        detector_type = 'FUV'
+
     photons_per_dn = DETECTOR_GAIN[detector_type]/DETECTOR_YIELD[detector_type]
     if data_unit == "DN":
         intensity_ph = photons_per_dn*convert_DN_to_photons(data, detector_type)
```
abit2: @DanRyanIrish is this line correct should it be multiplied by photons_per_dn after it has been converted to photons

DanRyanIrish:
Hi @abit2.  This looks good!  Could you add one more function to test_iris_tools.py to test that ```calculate_intensity_fractional_uncertainty()``` raises a ```ValueError``` when the wrong units are passed in.  You should be able to do it simply with:
```
with pytest.raise(ValueError):
    calculate_intensity_fractional_uncertainty(data, wrong_units, detector_type)
```

irispy/iris_tools.py
```diff
@@ -246,7 +246,7 @@ def get_iris_response(pre_launch=False, response_file=None, response_version=Non
 
 
 @custom_model
-def _gaussian1d_on_linear_bg(x, amplitude=None, mean=None, standard_deviation=None,
+def _gaussian1d_on_linear_bg(x=None, amplitude=None, mean=None, standard_deviation=None,
```
DanRyanIrish: I believe for @custom_model, ```x``` should be a positional argument, not a kwarg.  So the original version is actually correct.

abit2: temp11

abit2: @DanRyanIrish this happened when I tried this function

DanRyanIrish: This function should not be called directly.  It's only used as part of the fitting routine in ```_calculate_orbital_wavelength_variation()```.  So you don't need to test this line, if that's what you're trying to do.

DanRyanIrish:
Great stuff.  If this passes the Travis test, we should be ready to merge it!

abit2:
@DanRyanIrish any suggestions on what I should work next

PR #7, opened 2017-03-09
tests for IRISRasterClass

abit2:
ping @DanRyanIrish

### DIFF
--- a/irispy/data/test/__init__.py
+++ b/irispy/data/test/__init__.py
@@ -1,0 +1,44 @@
+"""Irispy test data files"""
+from __future__ import absolute_import
+
+import os
+import glob
+
+from astropy.utils.data import get_pkg_data_filename
+
+import irispy
+
+__all__ = ['rootdir', 'file_list', 'get_test_filepath']
+
+rootdir = os.path.join(os.path.dirname(irispy.__file__), "data", "test")
+
+def get_test_filepath(filename, **kwargs):
+    """
+    Return the full path to a test file in the ``data/test`` directory.
+
+    Parameters
+    ----------
+    filename : `str`
+        The name of the file inside the ``data/test`` directory.
+
+    Return
+    ------
+    filepath : `str`
+        The full path to the file.
+
+    See Also
+    --------
+
+    astropy.utils.data.get_pkg_data_filename : Get package data filename
+
+    Notes
+    -----
+
+    This is a wrapper around `astropy.utils.data.get_pkg_data_filename` which
+    sets the ``package`` kwarg to be 'irispy.data.test`.
+
+    """
+    return get_pkg_data_filename(filename, package="irispy.data.test", **kwargs)
+
+
+file_list = glob.glob(os.path.join(rootdir, '*.[!p]*'))
--- a/irispy/tests/test_spectrograph.py
+++ b/irispy/tests/test_spectrograph.py
@@ -33,8 +33,8 @@ def test_fits_data_comparison(iris_l2_test_raster):
 
     spectral_windows = iris_l2_test_raster.spectral_windows['name']
 
-    np.testing.assert_allclose(iris_l2_test_raster.data[spectral_windows[1]].data, data2)
     np.testing.assert_allclose(iris_l2_test_raster.data[spectral_windows[0]].data, data1)
+    np.testing.assert_allclose(iris_l2_test_raster.data[spectral_windows[1]].data, data2)
     np.testing.assert_allclose(iris_l2_test_raster.data[spectral_windows[2]].data, data3)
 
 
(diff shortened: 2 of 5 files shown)


DanRyanIrish:
Hi @abit2.  Thanks for another PR.  I'll take a look at this today.  Just a quick request.  When making PRs, can you give a brief explanation of what the PR does/changes and why?  Not just a ping.  Just makes it easier and quicker for me to figure out what's going on.  Feel free to include a ping as well though ;)  Cheers.

irispy/tests/test_spectrograph.py
```diff
@@ -1,2 +1,49 @@
...
+    np.testing.assert_allclose(iris_l2_test_raster.data[spectral_windows[2]].data, data1)
+    np.testing.assert_allclose(iris_l2_test_raster.data[spectral_windows[0]].data, data2)
+    np.testing.assert_allclose(iris_l2_test_raster.data[spectral_windows[1]].data, data3)
+
+
+def test_wcs(iris_l2_test_raster):
+    wcs_l = iris_l2_test_raster.wcs
+    hdulist = fits.open(os.path.join(testpath, 'iris_l2_20170222_153635_3690215148_raster_t000_r00000.fits'))
+    for key, value in six.iteritems(wcs_l):
+		for key_, value_ in wcs_l[key].iteritems():
```
DanRyanIrish: Inconsistent indentation here.

irispy/tests/test_spectrograph.py
```diff
@@ -1,2 +1,49 @@
...
+    hdulist = fits.open(os.path.join(testpath, 'iris_l2_20170222_153635_3690215148_raster_t000_r00000.fits'))
+    data1 = copy.deepcopy(hdulist[1].data)
+    data2 = copy.deepcopy(hdulist[2].data)
+    data3 = copy.deepcopy(hdulist[3].data)
+
+    data1[hdulist[1].data == -200.] = np.nan
+    data2[hdulist[2].data == -200.] = np.nan
+    data3[hdulist[3].data == -200.] = np.nan
+
+    spectral_windows = iris_l2_test_raster.data.keys()
```
DanRyanIrish: I think it would be better to get the spectral window names from ```iris_l2_test_raster.spectral_windows.name```.  This means we aren't locked into ```iris_l2_test_raster.data``` being a ```dict```.

irispy/data/test/__init__.py
```diff
@@ -0,0 +1,44 @@
...
+    -----
+
+    This is a wrapper around `astropy.utils.data.get_pkg_data_filename` which
+    sets the ``package`` kwarg to be 'irispy.data.test`.
+
+    """
+    return get_pkg_data_filename(filename, package="irispy.data.test", **kwargs)
+
+
+file_list = glob.glob(os.path.join(rootdir, '*.[!p]*'))
```
DanRyanIrish: Why is this variable defined?  Is it used anywhere?

abit2: not yet but it give the list of all the files in the directory

DanRyanIrish: What's the intention of including it in this PR?

DanRyanIrish:
There's a FileNotFoundError when Travis runs the the tests:
```FileNotFoundError: [Errno 2] No such file or directory: '/tmp/irispy-test-m8jwakk_/lib.linux-x86_64-3.5/irispy/data/test/iris_l2_20170222_153635_3690215148_raster_t000_r00000.fits'```
Looks like the filepath is not equalling what you think it is.  Perhaps the filepath should be defined in the test_spectrograph.py file in a more direct way?

abit2:
@DanRyanIrish it is passing if you run the test with
"py.test irispy/tests/test_spectrograph.py"
but not with "python setup.py test --coverage"

DanRyanIrish:
That makes me think the way you've defined the filepath isn't generally applicable.

abit2:
@DanRyanIrish 
it is same as the way in https://github.com/sunpy/sunpy/blob/master/sunpy/map/tests/test_mapbase.py

irispy/data/test/__init__.py
abit2: it can help in giving the list of files when using the terminal, maybe?

DanRyanIrish: Unless there's a clear reason we want it in, I think we should omit it in this PR.  If we find we need it for something later we can always add it in then.

DanRyanIrish:
Hmmm.  We might have to bring @Cadair to explain this then.  Or point us to who can explain it.

DanRyanIrish:
It all passes!  Good work @abit2 👍

PR #11, opened 2017-03-15
Copied mapcubed module from ehsteve's sunpy PR

DanRyanIrish:
Copied mapcubed module from ehsteve's sunpy PR so we can start work with it while the PR is being finished and merged into sunpy.

### DIFF
--- a/irispy/mapcubed.py
+++ b/irispy/mapcubed.py
@@ -1,0 +1,590 @@
+"""A Python MapCube Object"""
+from __future__ import absolute_import, division, print_function
+#pylint: disable=W0401,W0614,W0201,W0212,W0404
+
+from copy import deepcopy
+
+import numpy as np
+import matplotlib.animation
+
+from sunpy.map import GenericMap
+from sunpy.map.map_factory import Map
+from sunpy.visualization.mapcubeanimator import MapCubeAnimator
+from sunpy.visualization import wcsaxes_compat
+from sunpy.util import expand_list
+from sunpy.extern.six.moves import range
+import astropy.units as u
+from sunpy.time import parse_time
+from sunpy.lightcurve import LightCurve
+from pandas import DataFrame
+
+from sunpy import config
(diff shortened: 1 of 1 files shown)


abit2:
@DanRyanIrish can you please merge this so I can start working on SJIMapSequence

DanRyanIrish:
@abit2, yes I'll merge this today.  However, I would advise when working on SJIMapSequence to, as much as is possible, develop an abstract API to minimise how dependent we are on the specifics of MapCubed.  In light of the disagreements in riot last week, it is possible that we may want to use something other than MapCubed in the future if a better data cube object is developed in sunpy or astropy etc.  MapCubed is also bound to change before and after it is merged in sunpy.  So an abstract API will help minimise additional work for SJIMapSequence in the long run.

Cadair:
I maintain this was not the right way to go...

DanRyanIrish:
@Cadair, what do you suggest that we can use now to start developing API?

Cadair:
From my limited knowledge of the data I think that sunpy/cube is a closer fit to what you need.

DanRyanIrish:
So what's the point of MapCubed then?  As far as I know there isn't anything inherently different between AIA images and SJI images.  They are directly focused images of the Sun in EUV/UV.  Why would IRIS SJI need something inherently different than AIA?

Cadair:
> So what's the point of MapCubed then?

I am not really sure on this you would have to ask @ehsteve. As written MapCubed is a collection of 2D headers and one contiguous 3D array.

> Why would IRIS SJI need something inherently different than AIA?

AIA images are a collection of 2D images and metadata, whereas SJI is one 3D array and a singular 3D header. sunpy/cube was written specifically to address the case of a 3D (or more) array with a single corresponding header.

Again, all of this is as MapCubed is currently implemented. Although, I don't see the point in re-implementing something that looks a lot like sunpy/cube.

DanRyanIrish:
OK.  Thanks for the summary @Cadair.  You have a good point.

PR #14, opened 2017-04-11
sji movie tools

shelbej:
new sji movie tools ready for review

### DIFF
--- a/irispy/sji.py
+++ b/irispy/sji.py
@@ -106,74 +106,54 @@ def save_to_mp4(ani, outputfile, fps=60):
     print('Successfully saved to ' + outputfile)
 
 
-def dustbuster(filelist, clobber=False):
+def dustbuster(mc):
     """
-       Read SJI fits files and return Inpaint-corrected fits files.
-       Image inpainting involves filling in part of an image or video
-       using information from the surrounding area.
+           Read SJI fits files and return Inpaint-corrected fits files.
+           Image inpainting involves filling in part of an image or video
+           using information from the surrounding area.
 
-       Parameters
-       ----------
-       filelist: `str` or `list`
-           File to read, if single file string detected, will
-           create a list of length 1.
+           Parameters
+           ----------
+           mc: `sunpy.map.MapCube`
+               Mapcube to read
 
 
-       Returns
-       -------
-       NoneType
-           Inpaint-corrected fits to clobber input file
-       """
-
-    if type(filelist) == str:  # convert single file to list
-        filelist = [filelist]
-    nfile = 0
-    for fname in filelist:
-        nfile += 1
-        image_header = fits.getheader(fname)
-        image_orig = fits.getdata(fname)
-        nx = image_header.get('NRASTERP')
-        # TODO mapcube input
-
-        img_shape = image_orig.shape
-        ndx = img_shape[0]
-
-        image_result = []
+           Returns
+           -------
+           mc: `sunpy.map.MapCube`
+               Inpaint-corrected Mapcube
+               Optional, save inpaint-corrected fits file
+           """
+    image_result = []
+    ndx = len(mc)
+    for i, map in enumerate(mc):
+        image_orig = map.data
+        nx = map.meta.get('NRASTERP')
         firstpos = range(ndx)[0::nx]
+        #  Create mask with values < 1, excluding frame (-200)
+        mask = np.zeros(image_orig.shape)
+        mask[np.where(image_orig < 1)] = 1
+        mask[np.where(image_orig == -200)] = 0
+        image_fix = image_orig.copy()
+        if nx <= 50:  # sparse/coarse raster
+            skip = 1
+            secpos = [-1]
+            thirdpos = [-1]
+        elif nx > 50:  # dense raster
+            skip = 3
+            secpos = range(ndx)[1::nx]
+            thirdpos = range(ndx)[2::nx]
 
-        for i in range(0, ndx):
-            #  Create mask with values < 1, excluding frame (-200)
-            mask = np.zeros(image_orig[i].shape)
-            mask[np.where(image_orig[i] < 1)] = 1
-            mask[np.where(image_orig[i] == -200)] = 0
-
-            image_fix = image_orig[i].copy()
-            if nx <= 50:     # sparse/coarse raster
-                skip = 1
-                secpos = [-1]
-                thirdpos = [-1]
-            elif nx > 50:    # dense raster
-                skip = 3
-                secpos = range(ndx)[1::nx]
-                thirdpos = range(ndx)[2::nx]
-
-            if (i in firstpos) or (i in secpos) or (i in thirdpos):
-                image_inpaint = image_orig[i + skip].copy()  # grab next frame
-            else:
-                image_inpaint = image_orig[i - skip].copy()  # grab prev frame
-
-            # Inpaint mask onto image
-            for layer in range(image_fix.shape[-1]):
-                image_fix[np.where(mask)] = image_inpaint[np.where(mask)]
-
-            image_result.append(image_fix)  # add corrected image to new list
-
-
-        #  overwrite old file
-        if clobber == True:
-            outputfile = fname
+        if (i in firstpos) or (i in secpos) or (i in thirdpos):
+            image_inpaint = mc[i + skip].data.copy()  # grab next frame
         else:
-            outputfile = fname + '_dustbuster'
-        fits.writeto(outputfile, image_result, header=image_header,
-                     output_verify='fix', clobber=True)
-        print("Saved to " + outputfile)
+            image_inpaint = mc[i - skip].data.copy()  # grab prev frame
+
+        # Inpaint mask onto image
+        for layer in range(image_fix.shape[-1]):
+            image_fix[np.where(mask)] = image_inpaint[np.where(mask)]
+
+        image_result.append(image_fix)  # add corrected image to new list
+        map.data = image_fix
+
+    return mc
--- a/irispy/sji.py
+++ b/irispy/sji.py
@@ -1,18 +1,22 @@
 import datetime
-import matplotlib.pyplot as plt
-import matplotlib.animation as animation
 import sunpy.cm as cm
 import sunpy.instr.iris
 import sunpy.map
 import numpy as np
+import numpy.ma as ma
 import matplotlib.colors as colors
 from astropy.io import fits
+import sunpy.io
+import sunpy.time
+
+
 '''
 This module provides movie tools for level 2 IRIS SJI fits file
 '''
-__all__ = ["sji_fits_to_cube", "save_to_mp4", "dustbuster"]
 
-def sji_fits_to_cube(filelist, start=0, stop=None, skip=None, grid=False):
+__all__ = ['SJI_fits_to_cube','SJI_to_cube', 'dustbuster']
+
+def SJI_fits_to_cube(filelist, start=0, stop=None, skip=None):
     """
     Read SJI files and return a MapCube. Inherits
     sunpy.instr.iris.SJI_to_cube to stitch multiple sji fits
@@ -59,52 +63,51 @@ def sji_fits_to_cube(filelist, start=0, stop=None, skip=None, grid=False):
 
     #  todo: pointing correction(rot_hpc)
 
-    #  Option to overlay sun coordinates
-    if grid:
-        iris_cube[0].draw_grid(linestyle='--', color='orange')
 
     return iris_cube
 
-
-def save_to_mp4(ani, outputfile, fps=60):
-    '''
-    Set up ffmpeg writer
-    Save an animation of the MapCube
-    Example:
-    mc = sunpy.map.Map(files, cube=True)
-    ani = mc.plot()
-    save_to_mp4(ani, "myfile.mp4")
-
+def SJI_to_cube(filename, start=0, stop=None, hdu=0):
+    """
+    Read a SJI file and return a MapCube
+    .. warning::
+        This function is a very early beta and is not stable. Further work is
+        on going to improve SunPy IRIS support.
     Parameters
     ----------
-    ani:`matplotlib.animation.FuncAnimation`
-        Input animation
-    Outputfile: `str`
-        Path name of output file (.mp4)
-    fps: `int`
-        Desired frames per seconds (default=60)
+    filename: string
+        File to read
+    start: int
+        Temporal axis index to create MapCube from
+    stop: int
+        Temporal index to stop MapCube at
+    hdu: int
+        Choose hdu index
+    Returns
+    -------
+    iris_cube: sunpy.map.MapCube
+        A map cube of the SJI sequence
+    """
 
-    Return
-    -----------
-    NoneType
-        mp4 to outputfile path
-    '''
+    hdus = sunpy.io.read_file(filename)
+    # Get the time delta
+    time_range = sunpy.time.TimeRange(hdus[hdu][1]['STARTOBS'],
+                                      hdus[hdu][1]['ENDOBS'])
+    splits = time_range.split(hdus[hdu][0].shape[0])
 
-    plt.rcParams['animation.ffmpeg_path'] = '/usr/local/bin/ffmpeg'
-    writer = animation.FFMpegWriter(fps=fps, metadata=dict(artist='SunPy'),
-                                    bitrate=64000)
-    if not writer.isAvailable():
-        path = input('Set FFMpeg path: (ex. /usr/local/bin/ffmpeg)')
-        plt.rcParams['animation.ffmpeg_path'] = path
-    writer = animation.FFMpegWriter(fps=fps, metadata=dict(artist='SunPy'),
-                                    bitrate=64000)
-    if not writer.isAvailable():
-        print('Cannot find available FFMpeg Writer')
-        return
+    if not stop:
+        stop = len(splits)
 
-    ani.save(outputfile, writer=writer)
-    print('Successfully saved to ' + outputfile)
+    headers = [hdus[hdu][1]]*(stop-start)
+    datas = hdus[hdu][0][start:stop]
 
+    # Make the cube:
+    iris_cube = sunpy.map.Map(list(zip(datas, headers)), cube=True)
+    # Set the date/time
+
+    for i, m in enumerate(iris_cube):
+        m.meta['DATE-OBS'] = splits[i].center.isoformat()
+
+    return iris_cube
 
 def dustbuster(mc):
     """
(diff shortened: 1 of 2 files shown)


irispy/sji.py
```diff
@@ -0,0 +1,188 @@
...
+    for fname in filelist[0:]:
+        newmap = sunpy.instr.iris.SJI_to_cube(fname)
+        for frame in newmap[start:stop:skip]:
+            if frame.mean() < frame.max():
+                cmap = cm.get_cmap(frame._get_cmap_name())
+                vmax = frame.mean()+3.*frame.std()
+                frame.plot_settings['cmap'] = cmap
+                frame.plot_settings['norm'] = colors.LogNorm(1, vmax)
+                #  todo: iris_intscale
+                resultmapcube.maps.append(frame)
```
DanRyanIrish: ```resultmapcube``` is not defined.

irispy/sji.py
```diff
@@ -0,0 +1,188 @@
...
+    ax = plt.axes(zorder=0)
+
+    ani = mc.plot(ax)
+    ani.save(outputfile, writer=writer)
+    print('Successfully saved to ' + outputfile)
+
+
+def dustbuster(filelist):
+    """
+       Read SJI fits files and return Inpaint-corrected fits files.
```
DanRyanIrish: Can some additional information on what "Inpaint-corrected" means?

irispy/sji.py
```diff
@@ -0,0 +1,188 @@
...
+    stop: step
+        Temporal index to stop MapCube at
+
+    skip: step
+        Temporal index to skip over MapCube at
+
+
+    Returns
+    -------
+    iris_cube: sunpy.map.MapCube
```
DanRyanIrish: Data types in docstrings should be surrounded by `. e.g. `sunpy.map.MapCube`.  This enables then to be linked correctly in the documentation.  Same other parameters definition in this file.

irispy/sji.py
```diff
@@ -0,0 +1,188 @@
...
+    Assumes hdu index of 0.
+
+
+    Parameters
+    ----------
+    filelist: string or list
+        File to read, if single file string detected, will
+        create a list of length 1.
+
+    start: step
```
DanRyanIrish: What data type is "step"?

DanRyanIrish:
Hi @shelbej.  I had a quick look at your PR.  From the documentation, it looks like the dustbuster function actually overwrites the fits files.  Is this the standard practice is the IRIS team's IDL software?  Either way, I don't think overwriting the downloaded fits files is the best way to go.  Perhaps the dustbuster function should be applied to a MapCube of SJI images.  Or perhaps it could write out a new fits file suffixed by _dustbusted or allow the user to specify a filename so they can overwrite if they really want to.  But I we don't think we should get rid of the original data by default.  Unless this function takes a long time to run, I think it should be a function we apply to data already read into Python.

irispy/sji.py
Cadair: To add a little more technical to this when writing documentation in this format the right hand side of the colon should normally be a Python type, and should be in single back ticks. Assuming start, stop and skip here are supposed to be integers it should look like this:

```
start: `int`
```

irispy/sji.py
```diff
@@ -0,0 +1,188 @@
...
+    if type(filelist) == str:
+        filelist = [filelist]
+    iris_cube = sunpy.map.MapCube()
+    counter = 1
+    setup = sunpy.instr.iris.SJI_to_cube(filelist[0])
+    xcen = setup[0].meta.get('xcen')
+    ycen = setup[0].meta.get('ycen')
+    time = datetime.datetime.strptime(setup[0].meta.get('startobs'),
+                                      '%Y-%m-%d' + 'T' + '%H:%M:%S.%f')
+    print(xcen, ycen, time)
```
Cadair: Generally we try and minimise the use of print statements in functions. If you need to inform users of something important it is usual to raise a warning.

irispy/sji.py
```diff
@@ -0,0 +1,188 @@
...
+    iris_cube: sunpy.map.MapCube
+        A map cube of the SJI sequence
+    """
+
+    if type(filelist) == str:
+        filelist = [filelist]
+    iris_cube = sunpy.map.MapCube()
+    counter = 1
+    setup = sunpy.instr.iris.SJI_to_cube(filelist[0])
+    xcen = setup[0].meta.get('xcen')
```
Cadair: are these used anywhere else in this function?

irispy/sji.py
```diff
@@ -0,0 +1,188 @@
...
+    #  todo: pointing correction(rot_hpc)
+
+    #  Option to overlay sun coordinates
+    if grid:
+        iris_cube[0].draw_grid(linestyle='--', color='orange')
+
+    return iris_cube
+
+
+def save_to_mp4(mc, outputfile, fps=60):
```
Cadair: I would suggest refactoring this function so that instead of a mapcube it takes an animation object as input, so you would use it like this:
```python
mc = sunpy.map.Map(files, cube=True)
ani = mc.plot()
save_to_mp4(ani, "myfile.mp4")
```

irispy/sji.py
```diff
@@ -0,0 +1,188 @@
...
+                image_fix[np.where(mask)] = image_inpaint[np.where(mask)]
+
+            image_result.append(image_fix)  # add corrected image to new list
+            counter += 1
+            print(str(counter) + ' of ' + str(ndx))
+
+        #  overwrite old file
+        print(len(image_result))
+        outputfile = file
+        fits.writeto(outputfile, image_result, header=image_header,
```
Cadair: I suggest you modify this so it can optionally save to new files rather than overwriting.

Also I would make it return the list of files that it has modified or created.

irispy/sji.py
shelbej: should be iris_cube, fixed

irispy/sji.py
shelbej: added: 'Image inpainting involves filling in part of an image or video using information from the surrounding area.'

irispy/sji.py
shelbej: changed to 'int'

irispy/sji.py
shelbej: fixed

irispy/sji.py
shelbej: oh no, I was using those in another script for rot_hpc. removed

irispy/sji.py
shelbej: removed

irispy/sji.py
shelbej: I wrote a separate save function to set up the axes styles, otherwise ani.save() works just as well. Should I set up the axes in sji_fits_to_cube and just remove save_to_mp4?

irispy/sji.py
shelbej: Added a clobber keyword and changed the output file name to add '_dustbuster'. I should probably add a note to the header history too...

irispy/sji.py
```diff
@@ -0,0 +1,181 @@
...
+    print('Successfully saved to ' + outputfile)
+
+
+def dustbuster(filelist, clobber=False):
+    """
+       Read SJI fits files and return Inpaint-corrected fits files.
+       Image inpainting involves filling in part of an image or video
+       using information from the surrounding area.
+
+       **Warning** Will clobber input file.
```
DanRyanIrish: This warning should be removed and a description of the clobber keyword should be added to the Parameterts section of the docstring.

irispy/sji.py
```diff
@@ -0,0 +1,181 @@
...
+       Returns
+       -------
+       NoneType
+           Inpaint-corrected fits to clobber input file
+       """
+
+    if type(filelist) == str:  # convert single file to list
+        filelist = [filelist]
+    nfile = 0
+    for file in filelist:
```
DanRyanIrish: @Cadair can you confirm this?  Does the variable ```file``` have any specific meaning in python?  I've found it gets highlighted in Python-aware editors in a way different to other variable names (as here).  So have avoided using it, but never confirmed if this is good practice.  Is it better practice to use something like ```filename``` or have I been being a paranoid about using ```file```?

irispy/sji.py
```diff
@@ -0,0 +1,181 @@
...
+    plt.rcParams['ytick.direction'] = 'out'
+    ax = plt.axes(zorder=0)
+
+    ani = mc.plot(ax)
+    ani.save(outputfile, writer=writer)
+    print('Successfully saved to ' + outputfile)
+
+
+def dustbuster(filelist, clobber=False):
+    """
```
DanRyanIrish: I'm still unclear as to why this function operates on FITS files rather than on data itself, e.g. in a mapcube.  All the meta data is in the map.meta attribute.  In this implementation the user would then have to read in the dustbusted FITS file anyway.  Does it make more sense for this to take an sji mapcube object and return a new dustbusted sji mapcube?  The user can then write this out to a FITS file if they wanted.  Is there some other reasoning I don't see though?

irispy/sji.py
```diff
@@ -0,0 +1,181 @@
...
+        image_result = []
+        firstpos = range(ndx)[0::nx]
+
+        for i in range(0, ndx):
+            #  Create mask with values < 1, excluding frame (-200)
+            mask = np.zeros(image_orig[i].shape)
+            mask[np.where(image_orig[i] < 1)] = 1
+            mask[np.where(image_orig[i] == -200)] = 0
+
+            image_fix = image_orig[i].copy()
```
DanRyanIrish: Does .copy() create a shallow copy or a deepcopy?  And which one do you want?

irispy/sji.py
```diff
@@ -0,0 +1,181 @@
...
+            if nx <= 50:     # sparse/coarse raster
+                skip = 1
+            elif nx > 50:    # dense raster
+                skip = 3
+                secpos = range(ndx)[1::nx]
+                thirdpos = range(ndx)[2::nx]
+
+            if (i in firstpos) or (i in secpos) or (i in thirdpos):
```
DanRyanIrish: I tried running this function on the following file ```iris_l2_20141211_191222_3803105278_SJI_1400_t000.fits``` and got the following error at line 162: ```UnboundLocalError: local variable 'secpos' referenced before assignment```.

irispy/sji.py
shelbej: I was planning to have it accept either fits or mapcube, but it does take awhile to run so I wanted to be able to save the corrected image to reuse. Since we can't save and restore mapcubes, I prioritized the fits format. There is a todo for a mapcube input though

irispy/sji.py
shelbej: I believe its a shallow copy

irispy/sji.py
shelbej: There might not be value added here then. If ani is created outside save_to_mp4, then it might be better to let the user set up their own mpeg writer too. Once the writer is set up there's no difference from 'matplotlib.animation.FuncAnimation.save()'

irispy/sji.py
```diff
@@ -0,0 +1,181 @@
...
+    Read SJI files and return a MapCube. Inherits
+    sunpy.instr.iris.SJI_to_cube to stitch multiple sji fits
+    files stitched into one mapcube.
+    Also sets colormap and normalization
+    Assumes hdu index of 0.
+
+
+    Parameters
+    ----------
+    filelist: `string` or `list`
```
shelbej: fixed throughout

irispy/sji.py
shelbej: fixed

irispy/sji.py
shelbej: changed to fname

irispy/sji.py
shelbej: fixed

irispy/sji.py
shelbej: Oh I misunderstood, you meant fits input and map cube output? yes I could have it return a map cube, but like I said it does take a few min so I don't you would want to run it each time you read a file, which is why I initially clobbered the fits file.

irispy/sji.py
DanRyanIrish: So when you alter ```image_fix``` you also alter ```image_orig[i]```.  Is that what you want?

irispy/sji.py
DanRyanIrish: No I meant the input should be a mapcube.  If it takes a while to run, perhaps writing out the data to FITS files should be an option keyword by a default of False.  @Cadair have you got any opinion based on your intimate knowledge of the SunPy philosophy?

irispy/sji.py
shelbej: Yes

irispy/sji.py
Cadair: It depends how much you like an output FITS file to look like the input one. To my mind the best option might be to have a mapcube input and mapcube output with the option of calling `map.save()` for every map in the output mapcube as you dustbuster it.

DanRyanIrish: Yeh that would be my instinct too based on how we tend you use Python for solar physics analysis.  But I would also be happy for a an optional keyword to write out a FITS file similar in structure to the original rather than a separate file for each map.  But I think the default behaviour should be mapcube in, mapcube out.

DanRyanIrish:
@Cadair Can you interpret the Delphic oracle that is the Travis documentation build warnings log?  I don't see why it's failing for this PR.

irispy/sji.py
shelbej: Here's what I have for the map cube i/o and removed the save function for now. I tested it using a map cube constructed with sji_fits_to_cube.

irispy/sji.py
```diff
@@ -0,0 +1,159 @@
...
+        print('Cannot find available FFMpeg Writer')
+        return
+
+    ani.save(outputfile, writer=writer)
+    print('Successfully saved to ' + outputfile)
+
+
+def dustbuster(mc):
+    """
+           Read SJI fits files and return Inpaint-corrected fits files.
```
DanRyanIrish: Docstring should have same indentation as the opening triple quotes.

irispy/sji.py
```diff
@@ -0,0 +1,159 @@
...
+import sunpy.cm as cm
+import sunpy.instr.iris
+import sunpy.map
+import numpy as np
+import matplotlib.colors as colors
+from astropy.io import fits
+'''
+This module provides movie tools for level 2 IRIS SJI fits file
+'''
+
```
DanRyanIrish: Please add a line here defining the ```__all__``` list giving the functions to be documented online, i.e.
```__all__ = ["sji_fits_to_cube", "save_to_mp4", "dustbuster"]```

File: irispy/sji.py [hunk 17]
```diff
@@ -0,0 +1,159 @@
...
+    fps: `int`
+        Desired frames per seconds (default=60)
+
+    Return
+    -----------
+    NoneType
+        mp4 to outputfile path
+    '''
+
+    plt.rcParams['animation.ffmpeg_path'] = '/usr/local/bin/ffmpeg'
```
DanRyanIrish: When I tried this function, it failed here because it couldn't find the ffmpeg writer.  How should this be set/installed/etc.?

DanRyanIrish:
Hi @shelbej.  When I did a peek of the dustbusted mapcube I found that some dust regions remained uncorrected on some frames.  Is this expected?  Is it just an artifact in the animation and not in the maps themselves?  Can you try that you and see if you think it's normal?

irispy/sji.py
shelbej: It can installed with pip or anaconda, or also downloaded at ffmpeg.org

Cadair: yeah you can install it from conda but not pip.

irispy/sji.py
shelbej: added

irispy/sji.py
shelbej: I'm not married to ffmpeg, I just haven't used other writers. It seems to be the most common writer for matplotlib, but if someone knows a better alternative...

irispy/sji.py
shelbej: fixed

shelbej:
@DanRyanIrish were the uncorrected regions in the center of the image or close to the edges? Sometimes it can't find data to fill the holes around the edges.

irispy/sji.py
```diff
@@ -0,0 +1,158 @@
...
+    """
+    image_result = []
+    ndx = len(mc)
+    for i, map in enumerate(mc):
+        image_orig = map.data
+        nx = map.meta.get('NRASTERP')
+        firstpos = range(ndx)[0::nx]
+        #  Create mask with values < 1, excluding frame (-200)
+        mask = np.zeros(image_orig.shape)
+        mask[np.where(image_orig < 1)] = 1
```
ehsteve: could be just `mask[mask < 1]`

irispy/sji.py
```diff
@@ -0,0 +1,158 @@
...
+        Temporal index to skip over MapCube at
+
+
+    Returns
+    -------
+    iris_cube: `sunpy.map.MapCube`
+        A map cube of the SJI sequence
+    """
+
+    if type(filelist) == str:
```
Cadair:
```python
if isinstance(filelist, str):
```

irispy/sji.py
```diff
@@ -0,0 +1,158 @@
...
+        filelist = [filelist]
+    iris_cube = sunpy.map.MapCube()
+    for fname in filelist[0:]:
+        newmap = sunpy.instr.iris.SJI_to_cube(fname)
```
ehsteve: SJI_to_cube should be moved into this file as a private function.

irispy/sji.py
```diff
@@ -0,0 +1,158 @@
...
+                vmax = frame.mean()+3.*frame.std()
+                frame.plot_settings['cmap'] = cmap
+                frame.plot_settings['norm'] = colors.LogNorm(1, vmax)
+                #  todo: iris_intscale
+                iris_cube.maps.append(frame)
+
+    #  todo: pointing correction(rot_hpc)
+
+    #  Option to overlay sun coordinates
+    if grid:
```
Cadair: This shouldn't be here? this function has nothing to do with plotting?

irispy/sji.py
Cadair: I think this could be merged like this and then we can refactor it later. I think what you could do is load all the files first and then create the one mapcube. Rather than appending the maps to an empty cube.

irispy/sji.py
shelbej: I changed it to use numpy mask array, which returns a boolean array. It works, however I get this warning  when trying to inpaint the new values:

VisibleDeprecationWarning: using a boolean instead of an integer will result in an error in the future
Should I use  np.where(m.mask) here? 
        for layer in range(image_orig.shape[-1]):
            image_orig[m.mask] = image_inpaint[m.mask]

irispy/sji.py
shelbej: fixed

irispy/sji.py
shelbej: fixed

shelbej:
I ran into an error in dustbuster with different sized maps. Is there a way to crop/resize or remove maps? 
if not mc.all_maps_same_shape():
     do something that makes mc.all_maps_same_shape() = True

irispy/sji_test.py
```diff
@@ -0,0 +1,36 @@
+import sji
+import matplotlib.pyplot as plt
+import numpy.ma as ma
+
+iris_dir = '/Users/shelbe/Documents/IRIS/data/AR12641/'
+fname = [iris_dir + 'iris_l2_20170228_053417_3600106076_SJI_1330_t000.fits',
```
ehsteve: these should obviously not be hard coded. Has there been any discussion about how to handle test data?

ehsteve:
@shelbej just wanting on your fix before merging.

PR #17, opened 2017-05-12
Sample test data support

ehsteve:
This PR adds the ability to download sample data. It makes use of the sunpy sample data directory. This PR depends on SunPy [PR2116](https://github.com/sunpy/sunpy/pull/2116) which must be accepted before it works here.


irispy/data/sample.py
```diff
@@ -0,0 +1,21 @@
...
+_sampledata_dir = _config.get("downloads", "sample_dir")
+
+for _key in _sample_files:
+    if os.path.isfile(os.path.join(_sampledata_dir, _sample_files[_key])):
+        setattr(sys.modules[__name__], _key, os.path.join(_sampledata_dir, _sample_files[_key]))
+    else:
+        raise ImportError("Sample data file(s) missing. Use sunpy.data.download_sample_data() to get them.")
+
+file_dict = _sample_files
+file_list = list(_sample_files.values())
```
DanRyanIrish: What's the purpose of defining ```file_dict``` and ```file_list```?

irispy/data/_sample.py
```diff
@@ -0,0 +1,91 @@
...
+          "SJI_CUBE_1400": ("iris_l2_20170502_052551_3893010094_SJI_1400_t000.fits", ""),
+          "RASTER": ("iris_l2_20170502_052551_3893010094_raster_t000_r00000.fits", ".tar.gz")
+         }
+
+sample_files = {}
+for key in _files:
+    sample_files[key] = os.path.abspath(os.path.join(sampledata_dir, _files[key][0]))
+
+
+def download_data(progress=True, overwrite=True, timeout=None):
```
DanRyanIrish: Is this intended to be equivalent to ```sunpy.data._sample.download_sample_data```?

irispy/data/_sample.py
```diff
@@ -0,0 +1,91 @@
...
+        `astropy.utils.data.Conf.remote_timeout`.
+    Returns
+    -------
+    None
+    """
+
+    print(sampledata_dir)
+    number_of_files_fetched = 0
+    print("Downloading sample files to {}".format(sampledata_dir))
+    for handle, file_name in _files.items():
```
DanRyanIrish: If ```_files``` and ```_base_urls``` are variables defined at the top of the module for use in subsequent functions without further definition, should they not be capitalised?  It took me a while to figure out how they were defined.

irispy/data/_sample.py
ehsteve: yes

irispy/data/_sample.py
ehsteve: good point. fixed.

irispy/data/sample.py
ehsteve: they provides shortcuts through the next line like sunpy.AIA_171_IMAGE

Cadair:
Would it not be better if we could use the sunpy machinery for this, and just give it an extra set of URLs? or even just put the iris data in the sunpy sampledata?

ehsteve:
@Cadair i don't think that it should be part of the sunpy sample data as it's not really a solution for other affiliated packages. If the sunpy machinery could be made to be more easily used by affiliated packages that would nice.

irispy/data/_sample.py
```diff
@@ -0,0 +1,97 @@
...
+          "SJI_CUBE_1400": ("iris_l2_20170502_052551_3893010094_SJI_1400_t000.fits", ""),
+          "RASTER": ("iris_l2_20170502_052551_3893010094_raster.fits", ".tar.zip")
+          }
+
+sample_files = {}
+for key in _FILES:
+    sample_files[key] = os.path.abspath(os.path.join(sampledata_dir, _FILES[key][0]))
+
+
+def download_data(progress=True, overwrite=True, timeout=None):
```
DanRyanIrish: Should we not then call this ```download_sample_data``` to be consistent?  Or is there a reason you've named it differently?

DanRyanIrish:
@ehsteve Could you look into why the checks fail?  The test coverage one fails, I assume, because the overall test coverage has gone down which is fine at this stage.  But there are other failures too.

Cadair:
@ehsteve yeah that's fair, we should have a way to make this pulggable though, copy and pasting code is not sustainable either.

ehsteve:
FYI, we've decided to make downloading sample data pluggable in sunpy. I've submitted a PR to sunpy to do that. I'll update this PR once that is done. Also we've seen hosted the iris sample data on github but to be served by AWS (see https://github.com/sunpy/sample-data).

DanRyanIrish:
OK sounds good!

ehsteve:
Updated for new pluggable sunpy download architecture.

ehsteve:
[PR2116](https://github.com/sunpy/sunpy/pull/2116) has been merged so please update your SunPy and this PR should now work.

PR #19, opened 2017-05-22
[WIP] Added sjimap and sjicube first implementation

ehsteve:
A demo of this functionality is shown in this [jupiter notebook](https://github.com/ehsteve/jupyter-notebooks/blob/master/irispy_sjicube_demo.ipynb).


irispy/sji.py
```diff
@@ -1,20 +1,686 @@
...
+
+            self._meta = metas
+        elif len(input) > 1:
+            self.data = input[0]
+            self._meta = input[1]
+
+        self.ref_index = 0
+
+    def _get_map(self, index):
+        return SJIMap(self.data[index, :, :], self._meta[index], mask=self.data[index, :, :].mask)
```
wafels: I think you don't need the "mask=self.data[index, :, :].mask".

wafels: Try removing the "mask=self.data[index, :, :].mask" ?

tiagopereira:
Tried out this PR today. Seems really neat! I noticed a few issues:
- Loads the whole data into memory. This will be an issue with `astropy.io.fits` not allowing memmaps with `BSCALE` FITS files. Without loading into memory we won't have right DN units, but it will be much slower and possibly clog old machines. Not sure what the way forward is. Maybe give an option for memmap?
- I loaded an SJI with 670 frames. Then I did `sji.plot()`  and it proceeded to plot each after the other. Took more than 30 minutes on a fast machine, and I couldn't stop it without killing python. Perhaps some warning or check should be in order? 
- `.plot()` does not take roll into account

ehsteve:
@tiagopereira, thanks for taking a look at this PR! @Cadair, can you remind me what this problem is with memory mapping and pyfits with `BSCALE` fits files?

ehsteve:
@tiagopereira could you test this change to see if it loads the data with memory mapping (or send me a file to check myself). Thanks. Plotting still needs to be fixed.

irispy/sji.py
```diff
@@ -1,20 +1,720 @@
...
+            self._meta = metas
+        elif len(input) > 1:
+            self.data = input[0]
+            self._meta = input[1]
+            number_of_images = self.data.shape[0]
+
+        norm = []
+        cmap = []
+        for i in range(number_of_images):
+            norm.append(deepcopy(self[i].plot_settings['norm']))
```
DanRyanIrish: Got the following error when initializing using a single SJI file string:
```
AttributeError                            Traceback (most recent call last)
<ipython-input-14-db6437a43c2d> in <module>()
----> 1 sjicube = SJICube(sji1330_file)

~/sunpy_dev/ehsteve_irispy/irispy/sji.py in __init__(self, input)
    182         cmap = []
    183         for i in range(number_of_images):
--> 184             norm.append(deepcopy(self[i].plot_settings['norm']))
    185             cmap.append(deepcopy(self[i].plot_settings['cmap']))
    186 

~/sunpy_dev/ehsteve_irispy/irispy/sji.py in __getitem__(self, key)
    196         work.  Otherwise, a mapcube is returned."""
    197         if isinstance(key, int):
--> 198             return self._get_map(key)
    199         elif isinstance(key, slice):
    200             return SJICube((self.data[key, :, :], self._meta[key]))

~/sunpy_dev/ehsteve_irispy/irispy/sji.py in _get_map(self, index)
    189 
    190     def _get_map(self, index):
--> 191         return SJIMap(self.data[index, :, :], self._meta[index], mask=self.data[index, :, :].mask)
    192 
    193     def __getitem__(self, key):

AttributeError: 'numpy.ndarray' object has no attribute 'mask'
```

ehsteve: fixed!

PR #22, opened 2017-06-05
  wiki implentation suggestions 

shelbej:
added Lijia's implentation suggestions


ehsteve:
hi @shelbej, looks like this PR includes a few code changes as well. Did you want to include those too?

Issue #18, opened 2017-05-22
Add information for masking of dust on SJI sensor


ehsteve:
This information should be added as a data file in the data directory of the package.

DanRyanIrish:
@BaptistePellorceAstro, @tiagopereira and I have been discussing how to implement this.  There is currently an open PR (#74).  In a telecon, we concluded that using the available dust mask files from the IRIS team are not suitable as they are a level 1 product and describe the whole CCD.  The data processing involved in producing level 2 from level 1 can include rotation, interpolation, and FOV changes.  This means the level 1 dust mask would have to processed similarly which is beyond the scope of IRISpy.

One way of moving ahead could be a more functional approach.  This has the advantage of being less RAM hungry and also means a separate dust mask does not have to added to the slicing.  Instead, only the ```self.mask``` is altered which is already handling by the classes' ```__getitem__``` methods:
* Create a function in ```iris_tools```, say ```calculate_dust_mask``` that takes in a data array and returns a boolean dust mask array.  It could define dust pixels as those with DN < 0.5 but not equal to -200, as this is the unexposed pixel value.  Dust pixels should be marked as ```True``` and non-dust pixels as ```False```.
* Create a method, say ```IRISMapCube.apply_dust_mask```, that calls the ```calculate_dust_mask``` and sets the values of the dust pixels in the data mask to ```True```, i.e. 
```python
dust_mask = calculate_dust_mask(self.data)
self.mask[dust_mask] = True
```
* This method could have an ```undo=False``` kwarg.  If ```undo=True```, ```IRISMapCube.apply_dust_mask``` could undo the application of the dust mask, i.e. 
```python
self.mask[dust_mask] = False
```

Does this sound like a viable way forward @BaptistePellorceAstro @tiagopereira ?

BaptistePellorceAstro:
Yes, I will work on that. In my reading function, I put nan value instead of -200 for unexposed pixels. Is the `calculate_dust_mask` can be used for other obersevation than IRIS ? If no, I can just keep my code.

DanRyanIrish:
@BaptistePellorceAstro, did you leave a comment on this issue?  I got an email with a comment but can't see it here.

BaptistePellorceAstro:
Yes, I get my answer just after commenting ;)

PR #51, opened 2017-08-24
Dustbuster and intscale correctors

shelbej:
intscale added and dustbuster updated to be compatible with the SJICube/SJIMap, adds dustmask attribute to SJIMap https://github.com/sunpy/irispy/issues/18


irispy/sji.py
```diff
@@ -839,30 +850,34 @@ def dustbuster(mc):
     """
     image_result = []
     ndx = len(mc)
+    #intscale(mc)
+
     for i, map in enumerate(mc):
```
wafels: map is reserved word in Python.  it should not be used as a variable name.

PR #59, opened 2017-09-22
IRISSpectrogramSequence and IRISSpectrogram Classes

DanRyanIrish:
This PR creates a new class, ```IRISSpectrogram``` for IRIS spectral data described by one data cube and wcs object.  ```IRISSpectrogram``` is subclassed from ```NDCube``` but also contains methods for unit conversion, radiometric calibration, and exposure time correction.

The ```IRISSpectrogramSequence``` class has been refactored to use ```IRISSpectrogram``` and has also been given methods to perform unit conversion, radiometric calibration and exposure time correction to all spectrograms in the sequence.

Finally, ```IRISSpectrograph``` has been updated to used these classes and the metadata assignments have been slightly altered.


DanRyanIrish:
Hi @Cadair.  This PR has grown much bigger than I intended it to be.  I don't expect you to look at it all.  Could you just have a quick look at the ```IRISSpectrogram``` and ```IRISSpectrogramSequence``` classes in spectrograph.py and give any comments on API?
Cheers!

irispy/iris_tools.py
```diff
@@ -473,14 +395,351 @@ def _calculate_orbital_wavelength_variation(data_array, date_data_created, slit_
...
+    Returns
+    -------
+    new_data_arrays: `list` of `numpy.ndarray`s
+        Data arrays with exposure time correction removed.
+
+    new_unit_time_accounted: `astropy.unit.Unit`
+        Unit of new data arrays after exposure time correction removed.
+
+    """
+    if u.s not in (old_unit*u.s).decompose().bases:
```
Cadair: add comment to this.

irispy/iris_tools.py
```diff
@@ -473,14 +395,351 @@ def _calculate_orbital_wavelength_variation(data_array, date_data_created, slit_
...
+    -------
+    new_data_quantities: `list` of `astropy.units.Quantity`s
+        Data quantities converted to radiance or counts/s
+        depending on value of undo kwarg.
+
+    """
+    # Check data quantities are in the right units.
+    if undo is True:
+        for data in data_quantities:
+            assert data.unit.is_equivalent(RADIANCE_UNIT)
```
Cadair: it would be better to raise a proper error with a message rather than just using assert. (Same below)

irispy/iris_tools.py
```diff
@@ -473,14 +395,351 @@ def _calculate_orbital_wavelength_variation(data_array, date_data_created, slit_
...
+
+    Returns
+    -------
+    new_data_quantities: `list` of `astropy.units.Quantity`s
+        Data quantities converted to radiance or counts/s
+        depending on value of undo kwarg.
+
+    """
+    # Check data quantities are in the right units.
+    if undo is True:
```
dpshelio: `if undo`

irispy/spectrograph.py
```diff
@@ -208,6 +209,363 @@ def spectral_windows(self):
...
+
+        """
+        detector_type = iris_tools.get_detector_type(self.meta)
+        if new_unit_type == "radiance" or self.unit.is_equivalent(iris_tools.RADIANCE_UNIT):
+            # Get spectral dispersion per pixel.
+            spectral_wcs_index = np.where(np.array(self.wcs.wcs.ctype) == "WAVE")[0][0]
+            spectral_dispersion_per_pixel = self.wcs.wcs.cdelt[spectral_wcs_index] * \
+                                            self.wcs.wcs.cunit[spectral_wcs_index]
+            # Get solid angle from slit width for a pixel.
+            lat_wcs_index = np.where(np.array(self.wcs.wcs.ctype) == "HPLT-TAN")[0][0]
```
Cadair: would probably be good to only check for `HPLT` here not the projection as well.

irispy/spectrograph.py
```diff
@@ -208,6 +209,363 @@ def spectral_windows(self):
...
+
+        copy: `bool`
+            If True a new instance with the converted data values is return.
+            If False, the current instance is overwritten.
+            Default=False
+
+        """
+        converted_data_list = []
+        for cube in self.data:
+            converted_data_list.append(cube.to(new_unit_type))
```
dpshelio:
```python
converted_data_list = list(map(cube.to, self.data))
```

irispy/spectrograph.py
```diff
@@ -208,6 +209,363 @@ def spectral_windows(self):
...
+    def to(self, new_unit_type):
+        """
+        Converts data, unit and uncertainty attributes to new unit type.
+
+        The presence or absence of the exposure time correction is
+        preserved in the conversions.
+
+        Parameters
+        ----------
+        new_unit_type: `str`
```
Cadair: surely this can also be a unit object?

irispy/spectrograph.py
```diff
@@ -208,6 +209,363 @@ def spectral_windows(self):
...
+            _extra_coords_to_input_format(self._extra_coords, self.missing_axis),
+            mask=self.mask, missing_axis=self.missing_axis)
+
+def _get_meta_values(keys, meta, unknown_message="Unknown"):
+    values = []
+    for key in keys:
+        try:
+            value = meta[key]
+        except KeyError:
+            value = unknown_message
```
dpshelio: if `meta` is a  dictionary like, you don't need this `try/except` block but only:
```python
value = meta.get(key, unknown_message)
```

If you want to get a list with all these values in `keys`:
```python
values = [meta.get(key, unknown_message) for key in keys]
```

irispy/tests/test_iris_tools.py
```diff
@@ -1,129 +1,114 @@
...
+# Arrays of DN
+SOURCE_DATA_DN = np.array([[ 0.563,  1.132, -1.343], [-0.719,  1.441, 1.566]])
+SOURCE_DATA_DN_1 = np.array([[1, 2, 3], [4, 5, 6]])
+
+# Arrays relating SOURCE_DATA_DN to photons in NUV and FUV
+SOURCE_DATA_PHOTONS_NUV = np.array([[ 10.134, 20.376, -24.174], [-12.942, 25.938, 28.188]])
+SOURCE_DATA_PHOTONS_FUV = np.array([[ 2.252, 4.528, -5.372], [-2.876, 5.764, 6.264]])
+
+# Arrays relating SOURCE_DATA_DN_1 and photons in SJI
+SOURCE_DATA_PHOTONS_SJI_1 = np.array([[18, 36, 54], [72, 90, 108]])
```
dpshelio: Should these numbers be in a file that are read here? maybe that file could contain sublevels based on some metadata, like date or version.

irispy/spectrograph.py
```diff
@@ -208,6 +209,363 @@ def spectral_windows(self):
...
+        # Return new instance of IRISSpectrogram with correction applied/undone.
+        return IRISSpectrogram(
+            new_data_arrays[0], self.wcs, new_data_arrays[1], new_unit, self.meta,
+            _extra_coords_to_input_format(self._extra_coords, self.missing_axis),
+            mask=self.mask, missing_axis=self.missing_axis)
+
+def _get_meta_values(keys, meta, unknown_message="Unknown"):
+    values = []
+    for key in keys:
+        try:
```
Cadair: this can just be `meta.get(key, unknown_message)` you don't need the try block.

irispy/tests/test_iris_tools.py
dpshelio: Oh, this is a test... forget my comment.

irispy/spectrograph.py
Cadair: that's not quite enough
```python
from functools import partial

converted_data_list = list(map(partial(cube.to, new_unit_type), self.data))
```
Also does this really need the list cast?

PR #62, opened 2017-10-25
sji animation bug fix

shelbej:
.plot() and .peek() should now return cube animations
memmap can now be turned on/off
added a .percentile() function to maps for statisical analysis, similar to .max(), .mean(), etc

### DIFF
--- a/irispy/sji.py
+++ b/irispy/sji.py
@@ -852,32 +852,45 @@ def dustbuster(mc):
     mc: `sunpy.map.MapCube`
         Inpaint-corrected Mapcube
     """
-    image_result = []
-    ndx = len(mc)
-    for i, map in enumerate(mc):
-        image_orig = map.data
-        nx = map.meta.get('NRASTERP')
-        firstpos = range(ndx)[0::nx]
-        #  Create mask with values < 1, excluding frame (-200)
-        m = ma.masked_inside(image_orig, -199, .1)
+    intscale(mc)
+    image_inpaint = (mc.percentile(97).data) #(97th percentile to avoid SAA snow)
+    for i, m in enumerate(mc):
+        image_orig = m.data
 
-        if nx <= 50:  # sparse/coarse raster
-            skip = 1
-            secpos = [-1]
-            thirdpos = [-1]
-        elif nx > 50:  # dense raster
-            skip = 5
-            secpos = range(ndx)[1::nx]
-            thirdpos = range(ndx)[2::nx]
+        # Create mask with values < 10)
+        dustmask = ma.masked_inside(image_orig,-199,0)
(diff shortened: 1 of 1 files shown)


DanRyanIrish:
Hi @shelbej.  How's this PR coming along?

irispy/sji.py
```diff
@@ -502,16 +515,16 @@ def annotate_frame(i):
             axes.set_title("{s.name}".format(s=self[i]))
 
             # x-axis label
-            if self[0].coordinate_system.x == 'HG':
-                xlabel = 'Longitude [{lon}'.format(lon=self[i].spatial_units.x)
+            if self[0].coordinate_system.axis1 == 'HG':
```
Cadair: This should be re-written to use this: https://github.com/sunpy/sunpy/blob/master/sunpy/visualization/visualization.py#L38 helper function from SunPy. As map now uses here: https://github.com/sunpy/sunpy/blob/master/sunpy/map/mapbase.py#L1720

irispy/sji.py
```diff
@@ -539,9 +552,10 @@ def updatefig(i, im, annotate, ani_data, removes):
             # norm.autoscale_None(ani_data[i].data)
             im.set_norm(norm)
 
-            if wcsaxes_compat.is_wcsaxes(axes):
+            if wcsaxes_compat.is_wcsaxes(im.axes):
                 im.axes.reset_wcs(self[i].wcs)
-                wcsaxes_compat.default_wcs_grid(axes)
+                wcsaxes_compat.default_wcs_grid(im.axes, self[i].spatial_units,
+                                             self[i].coordinate_system)
```
Cadair: indentation is wrong

irispy/sji.py
shelbej: fixed

irispy/sji.py
shelbej: fixed

irispy/sji.py
```diff
@@ -135,20 +135,24 @@ class SJICube(object):
...
 
-    def __init__(self, input):
+    def __init__(self, input, memmap=False):
         """Creates a new instance"""
         if isinstance(input, str):
-            fits = pyfits.open(input, memmap=True, do_not_scale_image_data=True)
-            # TODO find a new masking value for unscaled data
-            #self.data = np.ma.masked_less_equal(fits[0].data, 0)
+            hdulist = fits.open(input, memmap=memmap, do_not_scale_image_data=memmap)
             self.data = fits[0].data
```
DanRyanIrish: This line errors.  I think you mean ```self.data = hdulist[0].data```

shelbej: fixed

PR #68, opened 2018-04-04
Adding SJICube class using NDCube

BaptistePellorceAstro:
The first version of the new SJICube class using NDCube. Most of this class comes from the old SJICube class.

### DIFF
--- a/irispy/iris_tools.py
+++ b/irispy/iris_tools.py
@@ -27,33 +27,41 @@ DETECTOR_YIELD = {"NUV": 1., "FUV": 1.5, "SJI": 1.}
 SJI_DEFAULT_BSCALE = 0.25
 SJI_DEFAULT_BZERO = 7992.0
 DN_UNIT = {
-    "NUV": u.def_unit("DN_IRIS_NUV", DETECTOR_GAIN["NUV"]/DETECTOR_YIELD["NUV"]*u.photon),
-    "FUV": u.def_unit("DN_IRIS_FUV", DETECTOR_GAIN["FUV"]/DETECTOR_YIELD["FUV"]*u.photon),
-    "SJI": u.def_unit("DN_IRIS_SJI", DETECTOR_GAIN["SJI"]/DETECTOR_YIELD["SJI"]*u.photon),
-    "SJI_UNSCALED": u.def_unit("DN_IRIS_SJI_UNSCALED",
-                               (DETECTOR_GAIN["SJI"]/DETECTOR_YIELD["SJI"]*SJI_DEFAULT_BSCALE
-                                + SJI_DEFAULT_BZERO)*u.photon)}
-SJI_SCALING = ((DN_UNIT["SJI"],
+    "NUV": u.def_unit("DN_IRIS_NUV",
+                      DETECTOR_GAIN["NUV"] / DETECTOR_YIELD["NUV"]*u.photon),
+    "FUV": u.def_unit("DN_IRIS_FUV",
+                      DETECTOR_GAIN["FUV"]/DETECTOR_YIELD["FUV"]*u.photon),
+    "SJI": u.def_unit("DN_IRIS_SJI",
+                      DETECTOR_GAIN["SJI"]/DETECTOR_YIELD["SJI"]*u.photon),
+    "SJI_UNSCALED": u.def_unit("DN_IRIS_SJI_UNSCALED", u.photon)}
+# Define an equivalency between SJI and SJI_UNSCALED units
+SJI_SCALING = [(DN_UNIT["SJI"],
                 DN_UNIT["SJI_UNSCALED"],
                 lambda x: (x - SJI_DEFAULT_BZERO) / SJI_DEFAULT_BSCALE,
-                lambda x: x * SJI_DEFAULT_BSCALE + SJI_DEFAULT_BZERO))
+                lambda x: x * SJI_DEFAULT_BSCALE + SJI_DEFAULT_BZERO)]
 
-READOUT_NOISE = {"NUV": 1.2*DN_UNIT["NUV"], "FUV": 3.1*DN_UNIT["FUV"],
-                 "SJI": 1.2*DN_UNIT["SJI"], "SJI_UNSCALED" : 1.2*DN_UNIT["SJI_UNSCALED"]}
+READOUT_NOISE = {"NUV": 1.2*DN_UNIT["NUV"],
+                 "FUV": 3.1*DN_UNIT["FUV"],
+                 "SJI": 1.2*DN_UNIT["SJI"],
+                 "SJI_UNSCALED" : 1.2*DN_UNIT["SJI_UNSCALED"]}
 RADIANCE_UNIT = u.erg / u.cm ** 2 / u.s / u.steradian / u.Angstrom
 SLIT_WIDTH = 0.33*u.arcsec
 
 IRIS_RESPONSE_REMOTE_PATH = "https://sohowww.nascom.nasa.gov/solarsoft/iris/response/"
-RESPONSE_VERSION_FILENAMES = {"1": "iris_sra_20130211.geny", "2": "iris_sra_20130715.geny",
-                              "3": "iris_sra_c_20150331.geny", "4": "iris_sra_c_20161022.geny"}
+RESPONSE_VERSION_FILENAMES = {"1": "iris_sra_20130211.geny",
+                              "2": "iris_sra_20130715.geny",
+                              "3": "iris_sra_c_20150331.geny",
+                              "4": "iris_sra_c_20161022.geny"}
 
 # Define some custom error messages.
-APPLY_EXPOSURE_TIME_ERROR = "Exposure time correction has probably already " + \
-    "been applied since the unit already includes inverse time.  " + \
-    "To apply exposure time correction anyway, set 'force' kwarg to True."
-UNDO_EXPOSURE_TIME_ERROR = "Exposure time correction has probably already " + \
-    "been undone since the unit does not include inverse time.  " + \
-    "To undo exposure time correction anyway, set 'force' kwarg to True."
+APPLY_EXPOSURE_TIME_ERROR = ("Exposure time correction has probably already "
+                             "been applied since the unit already includes "
+                             "inverse time. To apply exposure time correction "
+                             "anyway, set 'force' kwarg to True.")
+UNDO_EXPOSURE_TIME_ERROR = ("Exposure time correction has probably already "
+                            "been undone since the unit does not include "
+                            "inverse time. To undo exposure time correction "
+                            "anyway, set 'force' kwarg to True.")
 
 # Define whether IRIS WCS is 0 or 1 origin based.
 WCS_ORIGIN = 1
--- a/irispy/iris_tools.py
+++ b/irispy/iris_tools.py
@@ -24,14 +24,20 @@ from ndcube import NDCube
 # paper.
 DETECTOR_GAIN = {"NUV": 18., "FUV": 6., "SJI": 18.}
 DETECTOR_YIELD = {"NUV": 1., "FUV": 1.5, "SJI": 1.}
-BSCALE = 0.25
-BZERO = 7992
+SJI_DEFAULT_BSCALE = 0.25
+SJI_DEFAULT_BZERO = 7992.0
 DN_UNIT = {
     "NUV": u.def_unit("DN_IRIS_NUV", DETECTOR_GAIN["NUV"]/DETECTOR_YIELD["NUV"]*u.photon),
     "FUV": u.def_unit("DN_IRIS_FUV", DETECTOR_GAIN["FUV"]/DETECTOR_YIELD["FUV"]*u.photon),
     "SJI": u.def_unit("DN_IRIS_SJI", DETECTOR_GAIN["SJI"]/DETECTOR_YIELD["SJI"]*u.photon),
     "SJI_UNSCALED": u.def_unit("DN_IRIS_SJI_UNSCALED",
-                               (DETECTOR_GAIN["SJI"]/DETECTOR_YIELD["SJI"]*BSCALE + BZERO)*u.photon)}
+                               (DETECTOR_GAIN["SJI"]/DETECTOR_YIELD["SJI"]*SJI_DEFAULT_BSCALE
+                                + SJI_DEFAULT_BZERO)*u.photon)}
+SJI_SCALING = ((DN_UNIT["SJI"],
+                DN_UNIT["SJI_UNSCALED"],
+                lambda x: (x - SJI_DEFAULT_BZERO) / SJI_DEFAULT_BSCALE,
+                lambda x: x * SJI_DEFAULT_BSCALE + SJI_DEFAULT_BZERO))
+
 READOUT_NOISE = {"NUV": 1.2*DN_UNIT["NUV"], "FUV": 3.1*DN_UNIT["FUV"],
                  "SJI": 1.2*DN_UNIT["SJI"], "SJI_UNSCALED" : 1.2*DN_UNIT["SJI_UNSCALED"]}
 RADIANCE_UNIT = u.erg / u.cm ** 2 / u.s / u.steradian / u.Angstrom
--- a/irispy/iris_tools.py
+++ b/irispy/iris_tools.py
@@ -33,7 +33,7 @@ DN_UNIT = {
                       DETECTOR_GAIN["FUV"]/DETECTOR_YIELD["FUV"]*u.photon),
     "SJI": u.def_unit("DN_IRIS_SJI",
                       DETECTOR_GAIN["SJI"]/DETECTOR_YIELD["SJI"]*u.photon),
-    "SJI_UNSCALED": u.def_unit("DN_IRIS_SJI_UNSCALED")}
+    "SJI_UNSCALED": u.def_unit("DN_IRIS_SJI_UNSCALED", u.ct)}
 # Define an equivalency between SJI and SJI_UNSCALED units
 SJI_SCALING = [(DN_UNIT["SJI"],
                 DN_UNIT["SJI_UNSCALED"],
@@ -42,8 +42,7 @@ SJI_SCALING = [(DN_UNIT["SJI"],
 
 READOUT_NOISE = {"NUV": 1.2*DN_UNIT["NUV"],
                  "FUV": 3.1*DN_UNIT["FUV"],
-                 "SJI": 1.2*DN_UNIT["SJI"],
-                 "SJI_UNSCALED" : 1.2*DN_UNIT["SJI_UNSCALED"]}
+                 "SJI": 1.2*DN_UNIT["SJI"]}
 RADIANCE_UNIT = u.erg / u.cm ** 2 / u.s / u.steradian / u.Angstrom
 SLIT_WIDTH = 0.33*u.arcsec
 
@@ -113,7 +112,6 @@ def get_iris_response(pre_launch=False, response_file=None, response_version=Non
     if response_file is not None:
         if not(os.path.isfile(response_file)):
             raise KeyError("Not a valid file path")
-
 
     # Ensure conflicting kwargs are not set.
     if response_file:
--- a/irispy/iris_tools.py
+++ b/irispy/iris_tools.py
@@ -33,7 +33,7 @@ DN_UNIT = {
                       DETECTOR_GAIN["FUV"]/DETECTOR_YIELD["FUV"]*u.photon),
     "SJI": u.def_unit("DN_IRIS_SJI",
                       DETECTOR_GAIN["SJI"]/DETECTOR_YIELD["SJI"]*u.photon),
-    "SJI_UNSCALED": u.def_unit("DN_IRIS_SJI_UNSCALED", u.photon)}
+    "SJI_UNSCALED": u.def_unit("DN_IRIS_SJI_UNSCALED")}
 # Define an equivalency between SJI and SJI_UNSCALED units
 SJI_SCALING = [(DN_UNIT["SJI"],
                 DN_UNIT["SJI_UNSCALED"],
--- a/irispy/iris_tools.py
+++ b/irispy/iris_tools.py
@@ -24,12 +24,16 @@ from ndcube import NDCube
 # paper.
 DETECTOR_GAIN = {"NUV": 18., "FUV": 6., "SJI": 18.}
 DETECTOR_YIELD = {"NUV": 1., "FUV": 1.5, "SJI": 1.}
+BSCALE = 0.25
+BZERO = 7992
 DN_UNIT = {
     "NUV": u.def_unit("DN_IRIS_NUV", DETECTOR_GAIN["NUV"]/DETECTOR_YIELD["NUV"]*u.photon),
     "FUV": u.def_unit("DN_IRIS_FUV", DETECTOR_GAIN["FUV"]/DETECTOR_YIELD["FUV"]*u.photon),
-    "SJI": u.def_unit("DN_IRIS_SJI", DETECTOR_GAIN["SJI"]/DETECTOR_YIELD["SJI"]*u.photon)}
+    "SJI": u.def_unit("DN_IRIS_SJI", DETECTOR_GAIN["SJI"]/DETECTOR_YIELD["SJI"]*u.photon),
+    "SJI_UNSCALED": u.def_unit("DN_IRIS_SJI_UNSCALED",
+                               (DETECTOR_GAIN["SJI"]/DETECTOR_YIELD["SJI"]*BSCALE + BZERO)*u.photon)}
 READOUT_NOISE = {"NUV": 1.2*DN_UNIT["NUV"], "FUV": 3.1*DN_UNIT["FUV"],
-                 "SJI": 1.2*DN_UNIT["SJI"]}
+                 "SJI": 1.2*DN_UNIT["SJI"], "SJI_UNSCALED" : 1.2*DN_UNIT["SJI_UNSCALED"]}
 RADIANCE_UNIT = u.erg / u.cm ** 2 / u.s / u.steradian / u.Angstrom
 SLIT_WIDTH = 0.33*u.arcsec
 
@@ -96,7 +100,7 @@ def get_iris_response(pre_launch=False, response_file=None, response_version=Non
         if not(os.path.isfile(response_file)):
             raise KeyError("Not a valid file path")
 
-    
+
     # Ensure conflicting kwargs are not set.
     if response_file:
         response_file_set = True
--- a/irispy/new_sji.py
+++ b/irispy/new_sji.py
@@ -15,10 +15,8 @@ from irispy import iris_tools
 
 __all__ = ['SJICube']
 
-TIME_FORMAT = config.get("general", "time_format")
-
 # the following value is only appropriate for byte scaled images
-BAD_PIXEL_VALUE = -200
+BAD_PIXEL_VALUE_SCALED = -200
 # the following value is only appropriate for unscaled images
 BAD_PIXEL_VALUE_UNSCALED = -32768
 
@@ -94,6 +92,13 @@ class SJICube(NDCube):
                          copy=copy, missing_axis=missing_axis)
 
     def __repr__(self):
+        #Conversion of the start date of OBS
+        date_start = self.meta.get("DATE_OBS", None)
+        date_start = date_start.isoformat() if date_start else None
+        #Conversion of the end date of OBS
+        date_end = self.meta.get("DATE_END", None)
+        date_end = date_end.isoformat() if date_end else None
+        #Representation of SJICube object
         return (
             """
     SJICube
@@ -101,10 +106,10 @@ class SJICube(NDCube):
     Observatory:\t\t {obs}
     Instrument:\t\t\t {instrument}
     Bandpass:\t\t\t {bandpass}
-    Obs. Start:\t\t\t {date_start:{tmf}}
-    Obs. End:\t\t\t {date_end:{tmf}}
-    Instance Start:\t\t {instance_start:{tmf}}
-    Instance End:\t\t {instance_end:{tmf}}
+    Obs. Start:\t\t\t {date_start}
+    Obs. End:\t\t\t {date_end}
+    Instance Start:\t\t {instance_start}
+    Instance End:\t\t {instance_end}
     Total Frames in Obs.:\t {frame_num}
     IRIS Obs. id:\t\t {obs_id}
     IRIS Obs. Description:\t {obs_desc}
@@ -113,15 +118,15 @@ class SJICube(NDCube):
     """.format(obs=self.meta.get('TELESCOP', None),
                instrument=self.meta.get('INSTRUME', None),
                bandpass=self.meta.get('TWAVE1', None),
-               date_start=self.meta["DATE_OBS"],
-               date_end=self.meta["DATE_END"],
-               instance_start=self.extra_coords["TIME"]["value"][0],
-               instance_end=self.extra_coords["TIME"]["value"][-1],
+               date_start=date_start,
+               date_end=date_end,
+               instance_start=self.extra_coords["TIME"]["value"][0].isoformat(),
+               instance_end=self.extra_coords["TIME"]["value"][-1].isoformat(),
                frame_num=self.meta.get("NBFRAMES", None),
                obs_id=self.meta.get('OBSID', None),
                obs_desc=self.meta.get('OBS_DESC', None),
                axis_types=self.world_axis_physical_types,
-               dimensions=self.dimensions, tmf=TIME_FORMAT))
+               dimensions=self.dimensions))
 
 
 def read_iris_sji_level2_fits(filename, memmap=False):
@@ -154,8 +159,8 @@ def read_iris_sji_level2_fits(filename, memmap=False):
         mask = data_nan_masked == BAD_PIXEL_VALUE_UNSCALED
         scaled = False
     elif not memmap:
-        data_nan_masked[data == BAD_PIXEL_VALUE] = np.nan
-        mask = data_nan_masked == BAD_PIXEL_VALUE
+        data_nan_masked[data == BAD_PIXEL_VALUE_SCALED] = np.nan
+        mask = data_nan_masked == BAD_PIXEL_VALUE_SCALED
         scaled = True
     # Derive unit and readout noise from detector.
     exposure_times = my_file[1].data[:, my_file[1].header["EXPTIMES"]]
@@ -187,17 +192,11 @@ def read_iris_sji_level2_fits(filename, memmap=False):
     date_obs = parse_time(date_obs) if date_obs else None
     date_end = my_file[0].header.get('DATE_END', None)
     date_end = parse_time(date_end) if date_end else None
-    startobs = my_file[0].header.get('STARTOBS', None)
-    startobs = parse_time(startobs) if startobs else None
-    endobs = my_file[0].header.get('ENDOBS', None)
-    endobs = parse_time(endobs) if endobs else None
     meta = {'TELESCOP': my_file[0].header.get('TELESCOP', None),
             'INSTRUME': my_file[0].header.get('INSTRUME', None),
             'TWAVE1': my_file[0].header.get('TWAVE1', None),
             'DATE_OBS': date_obs,
             'DATE_END': date_end,
-            'STARTOBS': startobs,
-            'ENDOBS': endobs,
             'NBFRAMES': my_file[0].data.shape[0],
             'OBSID': my_file[0].header.get('OBSID', None),
             'OBS_DESC': my_file[0].header.get('OBS_DESC', None)}
--- a/irispy/new_sji.py
+++ b/irispy/new_sji.py
@@ -1,0 +1,177 @@
+'''
+This module provides movie tools for level 2 IRIS SJI fits file
+'''
+
+from datetime import timedelta
+
+from ndcube import NDCube
+import numpy as np
+from astropy.io import fits
+import astropy.units as u
+from astropy.wcs import WCS
+from sunpy.time import parse_time
+from sunpy import config
+from irispy import iris_tools
+
+__all__ = ['SJICube']
+
+TIME_FORMAT = config.get("general", "time_format")
+
+# the following value is only appropriate for byte scaled images
+BAD_PIXEL_VALUE = -200
+# the following value is only appropriate for unscaled images
+BAD_PIXEL_VALUE_UNSCALED = -32768
+
+
+class SJICube(NDCube):
+    """
+    SJICube
+
+    Class representing SJI Images described by a single WCS
+
+    Parameters
+    ----------
+    data: `numpy.ndarray`
+        The array holding the actual data in this object.
+
+    wcs: `ndcube.wcs.wcs.WCS`
+        The WCS object containing the axes' information
+
+    unit : `astropy.unit.Unit` or `str`
+        Unit for the dataset.
+        Strings that can be converted to a Unit are allowed.
+
+    meta : dict-like object
+        Additional meta information about the dataset.
+
+    uncertainty : any type, optional
+        Uncertainty in the dataset. Should have an attribute uncertainty_type
+        that defines what kind of uncertainty is stored, for example "std"
+        for standard deviation or "var" for variance. A metaclass defining
+        such an interface is NDUncertainty - but isn’t mandatory. If the
+        uncertainty has no such attribute the uncertainty is stored as
+        UnknownUncertainty.
+        Defaults to None.
+
+    mask : any type, optional
+        Mask for the dataset. Masks should follow the numpy convention
+        that valid data points are marked by False and invalid ones with True.
+        Defaults to None.
+
+    extra_coords : iterable of `tuple`s, each with three entries
+        (`str`, `int`, `astropy.units.quantity` or array-like)
+        Gives the name, axis of data, and values of coordinates of a data axis
+        not included in the WCS object.
+
+    copy : `bool`, optional
+        Indicates whether to save the arguments as copy. True copies every
+        attribute before saving it while False tries to save every parameter
+        as reference. Note however that it is not always possible to save the
+        input as reference.
+        Default is False.
+
+    Examples
+    --------
+    >>> from irispy import sji
+    >>> from irispy.data import sample
+    >>> sji = read_iris_sji_level2_fits(sample.SJI_CUBE_1400)
+    """
+
+    def __init__(self, data, wcs, uncertainty=None, unit=None, meta=None,
+                 extra_coords=None, mask=None, copy=False, missing_axis=None):
+        """
+        Initialization of Slit Jaw Imager
+        """
+
+        # Initialize SJI_NDCube.
+        super(SJICube, self).__init__(data, wcs, uncertainty=uncertainty,
+                                      mask=mask, meta=meta, unit=unit,
+                                      extra_coords=extra_coords, copy=copy,
+                                      missing_axis=missing_axis)
+
+    def __repr__(self):
+        return (
+            """
+    SJICube
+    ---------
+    Observatory:\t\t {obs}
+    Instrument:\t\t\t {inst}
+    Bandpass:\t\t\t {wave}
+    Obs. Start:\t\t\t {date_start:{tmf}}
+    Obs. End:\t\t\t {date_end:{tmf}}
+    Num. of Frames:\t\t {frame_num}
+    IRIS Obs. id:\t\t {obs_id}
+    IRIS Obs. Description:\t {obs_desc}
+    Cube dimensions:\t\t {dimensions}
+    Axis Types:\t\t\t {axis_types}
+    """.format(obs=self.meta["TELESCOP"], inst=self.meta["INSTRUME"],
+               wave=self.meta["TWAVE1"], date_start=self.meta["DATE_OBS"],
+               date_end=self.meta["DATE_END"], frame_num=self.meta["NBFRAMES"],
+               obs_id=self.meta["OBSID"], obs_desc=self.meta["OBS_DESC"],
+               axis_types=self.world_axis_physical_types,
+               dimensions=self.dimensions, tmf=TIME_FORMAT))
+
+
+def read_iris_sji_level2_fits(filename):
+    """
+    Read IRIS level 2 SJI FITS from an OBS into an NDCube instance
+
+    Parameters
+    ----------
+    filename : `str`
+        File name to be read
+
+    Returns
+    -------
+    result: 'ndcube.NDCube'
+
+    """
+
+    # Open a fits file
+    my_file = fits.open(filename)
+    # Derive WCS, data and mask for NDCube from fits file.
+    wcs = WCS(my_file[0].header)
+    data = my_file[0].data
+    data_nan_masked = my_file[0].data
+    data_nan_masked[data == BAD_PIXEL_VALUE] = np.nan
+    mask = data_nan_masked == BAD_PIXEL_VALUE
+    # Derive unit and readout noise from detector.
+    exposure_times = my_file[1].data[:, my_file[1].header["EXPTIMES"]]
+    unit = iris_tools.DN_UNIT["SJI"]
+    readout_noise = iris_tools.READOUT_NOISE["SJI"]
+    # Derive uncertainty of data for NDCube from fits file.
+    uncertainty = u.Quantity(np.sqrt((data_nan_masked*unit).to(u.photon).value
+                                     + readout_noise.to(u.photon).value**2),
+                             unit=u.photon).to(unit).value
+    # Derive extra coordinates for NDCube from fits file.
+    times = np.array([parse_time(my_file[0].header["STARTOBS"])
+                      + timedelta(seconds=s)
+                      for s in my_file[1].data[:, my_file[1].header["TIME"]]])
+    pztx = my_file[1].data[:, my_file[1].header["PZTX"]] * u.arcsec
+    pzty = my_file[1].data[:, my_file[1].header["PZTY"]] * u.arcsec
+    xcenix = my_file[1].data[:, my_file[1].header["XCENIX"]] * u.arcsec
+    ycenix = my_file[1].data[:, my_file[1].header["YCENIX"]] * u.arcsec
+    obs_vrix = my_file[1].data[:, my_file[1].header["OBS_VRIX"]] * u.m/u.s
+    ophaseix = my_file[1].data[:, my_file[1].header["OPHASEIX"]]
+    extra_coords = [('TIME', 0, times), ("PZTX", 0, pztx), ("PZTY", 0, pzty),
+                    ("XCENIX", 0, xcenix), ("YCENIX", 0, ycenix),
+                    ("OBS_VRIX", 0, obs_vrix), ("OPHASEIX", 0, ophaseix),
+                    ("EXPOSURE TIME", 0, exposure_times)]
+    # Extraction of meta for NDCube from fits file.
+    date_obs = my_file[0].header.get('DATE_OBS', None)
+    date_obs = parse_time(date_obs) if date_obs else None
+    date_end = my_file[0].header.get('DATE_END', None)
+    date_end = parse_time(date_end) if date_end else None
+    meta = {'TELESCOP': my_file[0].header.get('TELESCOP'),
+            'INSTRUME': my_file[0].header.get('INSTRUME'),
+            'TWAVE1': my_file[0].header.get('TWAVE1'),
+            'DATE_OBS': date_obs,
+            'DATE_END': date_end,
+            'NBFRAMES': my_file[0].data.shape[0],
+            'OBSID': my_file[0].header.get('OBSID'),
+            'OBS_DESC': my_file[0].header.get('OBS_DESC')}
+
+    my_file.close()
+
+    return SJICube(data_nan_masked, wcs, uncertainty=uncertainty, unit=unit,
+                   meta=meta, mask=mask, extra_coords=extra_coords)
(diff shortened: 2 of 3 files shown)


irispy/sji.py
```diff
@@ -737,6 +738,94 @@ def base_difference(self, base=0, fraction=False):
...
+    ---------
+    Observatory:\t\t {obs}
+    Instrument:\t\t\t {inst}
+    Wavelength:\t\t\t {wave}
+    Obs. Start:\t\t\t {date_start:{tmf}}
+    Obs. End:\t\t\t {date_end:{tmf}}
+    Num. of Frames:\t\t {frame_num}
+    IRIS Obs. id:\t\t {obs_id}
+    IRIS Obs. Description:\t {obs_desc}
+    """.format(obs=self.meta["TELESCOP"], inst=self.meta["INSTRUME"],
```
DanRyanIrish: Also add the dimensions and world axis physical types of the cube at the end of the string, similar to what is done in the ```NDCube.__repr__```.  So something like:
```
Cube dimensions:\t {dimensions}
Axis Types:\t {axis_types}
```
where {dimensions} is ```_SJICube.dimensions``` and {axis_types} is ```_SJICube.world_axis_physical_types```.

Obviously check the number of ```\t``` each needs to align well.

irispy/sji.py
```diff
@@ -737,6 +738,94 @@ def base_difference(self, base=0, fraction=False):
         return SJICube(new_mc)
 
 
+class _SJICube(NDCube):
+
```
DanRyanIrish: Remove blank line here

DanRyanIrish:
@BaptistePellorceAstro: @Cadair suggested to me that we move your development into a new module, say ```new_sji.py``` to give us a clean slate and better manage the ```diff``` when it comes time to replace the old ```SJICube```.  I think it's a good idea.  Could you move ```_SJICube``` and ```read_iris_sji_level2_fits``` into such a new module as part of this PR?  It's just a cut-and-paste job.  It should be in the same directory as ```sji.py```.

irispy/sji.py
```diff
@@ -737,6 +738,94 @@ def base_difference(self, base=0, fraction=False):
...
+    Examples
+    --------
+    >>> from irispy import sji
+    >>> from irispy.data import sample
+    >>> sji = read_iris_sji_level2_fits(sample.SJI_CUBE_1400)
+    """
+
+    def __init__(self, data, wcs, uncertainty=None, unit=None, meta=None,
+                 extra_coords=None, mask=None, copy=False, missing_axis=None):
+
```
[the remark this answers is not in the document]
BaptistePellorceAstro: Removed

irispy/sji.py
```diff
@@ -737,6 +738,94 @@ def base_difference(self, base=0, fraction=False):
...
+                                       missing_axis=missing_axis)
+
+    def __repr__(self):
+        return (
+            """
+    SJICube
+    ---------
+    Observatory:\t\t {obs}
+    Instrument:\t\t\t {inst}
+    Wavelength:\t\t\t {wave}
```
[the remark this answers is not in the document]
BaptistePellorceAstro: Renamed

irispy/sji.py
BaptistePellorceAstro: Added

irispy/sji.py
BaptistePellorceAstro: Removed

irispy/new_sji.py
```diff
@@ -0,0 +1,177 @@
...
+                                      mask=mask, meta=meta, unit=unit,
+                                      extra_coords=extra_coords, copy=copy,
+                                      missing_axis=missing_axis)
+
+    def __repr__(self):
+        return (
+            """
+    SJICube
+    ---------
+    Observatory:\t\t {obs}
```
Cadair: only two tabs here but 3 on most of the other lines? Is that deliberate?

irispy/new_sji.py
```diff
@@ -0,0 +1,177 @@
...
+    """
+
+    def __init__(self, data, wcs, uncertainty=None, unit=None, meta=None,
+                 extra_coords=None, mask=None, copy=False, missing_axis=None):
+        """
+        Initialization of Slit Jaw Imager
+        """
+
+        # Initialize SJI_NDCube.
+        super(SJICube, self).__init__(data, wcs, uncertainty=uncertainty,
```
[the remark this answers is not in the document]
BaptistePellorceAstro: Ok thanks, I don't know that !

irispy/new_sji.py
BaptistePellorceAstro: Yes, this is to align all datas to have a better global view

irispy/new_sji.py
```diff
@@ -0,0 +1,177 @@
...
+        Default is False.
+
+    Examples
+    --------
+    >>> from irispy import sji
+    >>> from irispy.data import sample
+    >>> sji = read_iris_sji_level2_fits(sample.SJI_CUBE_1400)
+    """
+
+    def __init__(self, data, wcs, uncertainty=None, unit=None, meta=None,
```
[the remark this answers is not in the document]
BaptistePellorceAstro: I don't really understand what you mean by "define this method". For now __init__ is just creating the NDCube. 

If I just return the NDCube creation, I must write something else ?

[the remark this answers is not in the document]
Cadair: as written here, you don't need `__init__` at all. It will just inherit the one from NDCube. You only need to define the `__init__` here if you are putting more code in it than just the `super()` call.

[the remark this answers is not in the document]
BaptistePellorceAstro: Ok so you suggest me that I can remove this part as I don't do anything in the function ! Thanks for your help ! I think I'll let it for now if need but I'll remember it for my later work. Thanks !

BaptistePellorceAstro:
Hi, I always try what I post here to be sure. 
I have change the class name from `_SJICube` to `SJICube` to prevent the later insertion on sji. 
On my version, I have modified what Cadair told me to modify, I am just waiting for more changes to update the PR.

DanRyanIrish:
Very good policy :)  You don't have to wait for more changes.  You can update the PR and comments on lines that have changed will be marked as outdated (but still viewable) and other comments will remain.

irispy/new_sji.py
```diff
@@ -0,0 +1,177 @@
...
+    IRIS Obs. id:\t\t {obs_id}
+    IRIS Obs. Description:\t {obs_desc}
+    Cube dimensions:\t\t {dimensions}
+    Axis Types:\t\t\t {axis_types}
+    """.format(obs=self.meta["TELESCOP"], inst=self.meta["INSTRUME"],
+               wave=self.meta["TWAVE1"], date_start=self.meta["DATE_OBS"],
+               date_end=self.meta["DATE_END"], frame_num=self.meta["NBFRAMES"],
+               obs_id=self.meta["OBSID"], obs_desc=self.meta["OBS_DESC"],
+               axis_types=self.world_axis_physical_types,
+               dimensions=self.dimensions, tmf=TIME_FORMAT))
```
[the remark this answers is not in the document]
BaptistePellorceAstro: Nice, this is a good idea !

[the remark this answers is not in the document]
BaptistePellorceAstro: I think if we have a missing data, the error will come from the reading FITS function and not from __repr__ so I modified the meta data extraction process.

irispy/new_sji.py
```diff
@@ -0,0 +1,177 @@
...
+        return (
+            """
+    SJICube
+    ---------
+    Observatory:\t\t {obs}
+    Instrument:\t\t\t {inst}
+    Bandpass:\t\t\t {wave}
+    Obs. Start:\t\t\t {date_start:{tmf}}
+    Obs. End:\t\t\t {date_end:{tmf}}
+    Num. of Frames:\t\t {frame_num}
```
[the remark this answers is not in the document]
BaptistePellorceAstro: Renamed

irispy/new_sji.py
```diff
@@ -0,0 +1,177 @@
...
+    def __repr__(self):
+        return (
+            """
+    SJICube
+    ---------
+    Observatory:\t\t {obs}
+    Instrument:\t\t\t {inst}
+    Bandpass:\t\t\t {wave}
+    Obs. Start:\t\t\t {date_start:{tmf}}
+    Obs. End:\t\t\t {date_end:{tmf}}
```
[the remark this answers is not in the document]
BaptistePellorceAstro: Added but I don't know which data use for each key. I have DATE_OBS / DATE_END / STARTOBS / ENDOBS.

I saw that STARTOBS to ENDOBS is longer than DATE_OBS to DATE_END.

irispy/new_sji.py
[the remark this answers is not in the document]
DanRyanIrish: But what if a user creates an ```SJICube``` by not reading a FITS file?  For example if they have processed data in a CSV, Pickle, or custom FITS file?  All methods on ```SJICube``` should be independent of how the object is created.  Therefore, I think the ```self.meta.get``` implementation is best.  In fact it should also be implemented in the ```IRISSpectrograph.__repr__```.

irispy/new_sji.py
```diff
@@ -0,0 +1,190 @@
...
+    IRIS Obs. Description:\t {obs_desc}
+    Cube dimensions:\t\t {dimensions}
+    Axis Types:\t\t\t {axis_types}
+    """.format(obs=self.meta["TELESCOP"],
+               instrument=self.meta["INSTRUME"],
+               bandpass=self.meta["TWAVE1"],
+               date_start=self.meta["DATE_OBS"],
+               date_end=self.meta["DATE_END"],
+               instance_start=self.meta["STARTOBS"],
+               instance_end=self.meta["ENDOBS"],
```
DanRyanIrish: This should be 
```
instance_start=self.extra_coords["TIME"]["value"][0]
instance_end=self.extra_coords["TIME"]["value"][-1]
```

BaptistePellorceAstro:
_init_ function has been removed
FITS file opening use now memmap. I will make some changes to recreate a NaN mask.

irispy/new_sji.py
```diff
@@ -135,13 +124,13 @@ def read_iris_sji_level2_fits(filename):
     """
 
     # Open a fits file
-    my_file = fits.open(filename)
+    my_file = fits.open(filename, memmap=True, do_not_scale_image_data=True)
```
DanRyanIrish: The ```memmap``` kwarg also needs to be in the definition of the function so a user can turn it off an on.

I suggest the default should be ```False``` and then users can turn it on when they are dealing with dig data.  What do you think @tiagopereira ?

irispy/new_sji.py
```diff
@@ -135,13 +124,13 @@ def read_iris_sji_level2_fits(filename):
...
     # Derive WCS, data and mask for NDCube from fits file.
     wcs = WCS(my_file[0].header)
     data = my_file[0].data
     data_nan_masked = my_file[0].data
-    data_nan_masked[data == BAD_PIXEL_VALUE] = np.nan
-    mask = data_nan_masked == BAD_PIXEL_VALUE
+    data_nan_masked[data == BAD_PIXEL_VALUE_UNSCALED] = BAD_PIXEL_VALUE
+    mask = data_nan_masked == BAD_PIXEL_VALUE_UNSCALED
```
DanRyanIrish: There should be an if statement here checking whether ```memmap is True```.  If so, use ```BAD_PIXEL_VALUE_UNSCALED```, else use ```BAD_PIXEL_VALUE```.

```BAD_PIXEL_VALUE_UNSCALED``` also needs to be defined at the top of the file underneath where ``````BAD_PIXEL_VALUE``` is.

File: irispy/new_sji.py (hunk 8 above)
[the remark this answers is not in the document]
Comment (DanRyanIrish) [reply irispy/new_sji.py]: Don't forget this change too.  I see it was marked as outdated.

File: irispy/new_sji.py [hunk 14]
```diff
@@ -0,0 +1,179 @@
...
+    meta = {'TELESCOP': my_file[0].header.get('TELESCOP', "Unknown"),
+            'INSTRUME': my_file[0].header.get('INSTRUME', "Unknown"),
+            'TWAVE1': my_file[0].header.get('TWAVE1', "Unknown"),
+            'DATE_OBS': date_obs,
+            'DATE_END': date_end,
+            'STARTOBS': startobs,
+            'ENDOBS': endobs,
+            'NBFRAMES': my_file[0].data.shape[0],
+            'OBSID': my_file[0].header.get('OBSID', "Unknown"),
+            'OBS_DESC': my_file[0].header.get('OBS_DESC', "Unknown")}
```
DanRyanIrish: Having thought about this, ```my_file[0].header.get('OBSID', "Unknown")``` should be ```my_file[0].header.get('OBSID', None)```.   Same goes for all other occurences of ```"Unknown"```.  That way you can use if statements like:
```
if not self.meta["X"]:
    # Do something
````
rather than explicitly having to check whether == "Unknown".

File: irispy/new_sji.py [hunk 15]
```diff
@@ -0,0 +1,179 @@
...
+    Obs. Start:\t\t\t {date_start:{tmf}}
+    Obs. End:\t\t\t {date_end:{tmf}}
+    Instance Start:\t\t {instance_start:{tmf}}
+    Instance End:\t\t {instance_end:{tmf}}
+    Total Frames in Obs.:\t {frame_num}
+    IRIS Obs. id:\t\t {obs_id}
+    IRIS Obs. Description:\t {obs_desc}
+    Cube dimensions:\t\t {dimensions}
+    Axis Types:\t\t\t {axis_types}
+    """.format(obs=self.meta["TELESCOP"],
```
DanRyanIrish: Adding new reminder to change ```self.meta["X"]``` to ```self.meta.get("X", None)``` as the old comment was marked as outdated.

irispy/new_sji.py
```diff
@@ -0,0 +1,179 @@
...
+        as reference. Note however that it is not always possible to save the
+        input as reference.
+        Default is False.
+
+    Examples
+    --------
+    >>> from irispy import sji
+    >>> from irispy.data import sample
+    >>> sji = read_iris_sji_level2_fits(sample.SJI_CUBE_1400)
+    """
```
DanRyanIrish: If we are introducing the option to memory map (i.e. scale or not scale) the data, we should add a flag to state that.  So say, ```SJICube.scaled = True``` if memmap is ```False```, ```SJICube.scaled = False``` if memmap is ```True```, or ```SJICube.scaled = None``` if not specified.

This will require the ```SJICube.__init__``` to be put back as we are now doing something more than just the calling the ```NDCube.__init__```.  So the ```SJICube.__init__``` would look something like:
```
def __init__(self, data, wcs, uncertainty=None, unit=None, meta=None, mask=None, 
             extra_coords=None, copy=False, missing_axis=None, scaled=None):
    # Set whether SJI data is scaled or not.
    self.scaled = scaled
    # Call NDCube init
    super............
```
Then ```scaled``` must be set to ```True``` or ```False``` in ```read_iris_sji_level2_fits``` when it defines the ```SJICube``` object based on the value of the ```memmap``` keyword argument (kwarg).

Having the ```scaled``` attribute will enable a user to know how to interpret the data numbers even without knowing whether ```memmap``` was ```True``` or ```False``` during initialization.

We should also add an ```SJI_UNSCALED``` unit to the DN dictionary in ```iris_tools``` (https://github.com/sunpy/irispy/blob/master/irispy/iris_tools.py#L27) which is just the ```DN_UNIT["SJI"]/SJI_SCALE_FACTOR```.    (```SJI_SCALE_FACTOR``` should also be defined near the top of ```iris_tools.py```.)  Then based on whether ```memmap``` is ```True``` or ```False``` the ```unit``` of ```SJICube``` can be set to ```DN_UNIT["SJI"]``` or ```DN_UNIT["SJI_UNSCALED"]``` by ```read_iris_sji_level2_fits```.

irispy/new_sji.py
BaptistePellorceAstro: I wrote these lines to fix a bug before leaving the office, I will change it next week.

irispy/new_sji.py
BaptistePellorceAstro: Sure, this is better to use None if we'll create a test. I was using "Unknown" because this one is more relevant for a data.

irispy/new_sji.py
BaptistePellorceAstro: I did not change it because I was changing it on reading function, but now, with None, I have to modify it.

irispy/new_sji.py
```diff
@@ -0,0 +1,179 @@
...
+    my_file = fits.open(filename, memmap=True, do_not_scale_image_data=True)
+    # Derive WCS, data and mask for NDCube from fits file.
+    wcs = WCS(my_file[0].header)
+    data = my_file[0].data
+    data_nan_masked = my_file[0].data
+    data_nan_masked[data == BAD_PIXEL_VALUE_UNSCALED] = BAD_PIXEL_VALUE
+    mask = data_nan_masked == BAD_PIXEL_VALUE_UNSCALED
+    # Derive unit and readout noise from detector.
+    exposure_times = my_file[1].data[:, my_file[1].header["EXPTIMES"]]
+    unit = iris_tools.DN_UNIT["SJI"]
```
DanRyanIrish: if we have an ```SJICube.scaled``` attribute, this ```unit variable should be defined in an ```if...else...``` statement based on whether ```memmap``` is ```True``` or ```False```.

irispy/new_sji.py
DanRyanIrish: Perhaps the ```scaled``` argument should be a required argument, like ```data``` and ```wcs```.  Or perhaps it should default to ```True```.  Either of these ways will mean it will always have a Boolean value and will never be ```None```.  That way any infrastructure depending on it, e.g. ```unit``` will always be valid.  I'll give this is a little thought.

irispy/new_sji.py
```diff
@@ -0,0 +1,179 @@
...
+from irispy import iris_tools
+
+__all__ = ['SJICube']
+
+TIME_FORMAT = config.get("general", "time_format")
+
+# the following value is only appropriate for byte scaled images
+BAD_PIXEL_VALUE = -200
+# the following value is only appropriate for unscaled images
+BAD_PIXEL_VALUE_UNSCALED = -32768
```
DanRyanIrish: @tiagopereira Is the scaling always the same for SJI?  If so, we can hard code the ```BAD_PIXEL_VALUE_UNSCALED``` here.  Otherwise we'll have to think of a more flexible solution.

irispy/new_sji.py
BaptistePellorceAstro: memmap added to the definition of the function.
Default value of memmap turned to False

irispy/new_sji.py
BaptistePellorceAstro: None is now replacing "Unknown"

irispy/new_sji.py
BaptistePellorceAstro: If the header does not contain a value, the reading function will translate it to None. So we don't need to add this on repr.

BaptistePellorceAstro: I found an error on the dates reading. If date is None, repr can not display a value due to the time format. Do you have an idea to solve it ? 

I think I need to modify the time format in reading function but I don't know how...

irispy/new_sji.py
BaptistePellorceAstro: How can I define `SJI_SCALE_FACTOR` on `iris_tools.py` ?

`scaled` has been added on reading function and SJI class.

irispy/new_sji.py
BaptistePellorceAstro: `unit` statement has been added, I am waiting the `DN_UNIT["SJI_UNSCALED"]`

irispy/new_sji.py
DanRyanIrish: That's incorrect.  ```self.meta.get("X", None)``` is still needed here.  Otherwise you're assuming that the only way an ```SJICube``` will ever be created is via ```read_iris_sji_level2_fits```.  This is an unnecessarily restrictive assumption that means ```SJICube``` is not fully independent of ```read_iris_sji_level2_fits```, which of course was the whole point of writing a separate reading function in the first place.

DanRyanIrish: Could you provide more detail on the dates error?  I'm not sure what you mean.

irispy/new_sji.py
DanRyanIrish: Looks like ```"Unknown"``` is still being used here.  Perhaps you have not pushed your changes?

irispy/new_sji.py
DanRyanIrish: Firstly, could @tiagopereira chime in here and confirm whether the scaling is always the same for SJI data?  Depending on that answer, we can decide how to define it.

irispy/new_sji.py
BaptistePellorceAstro: I don't have pushed now, I am still working on some problems that I try to solve. I will push after.

irispy/new_sji.py
BaptistePellorceAstro: Sure, I forget the fact we can use other files ... sorry 

If I put `None` in dates (if there is no dates on fits file) as can do the reading function, my meta for time is `None` and `None` can't be in a time format with `tmf=TIME_FORMAT` and `TIME_FORMAT = config.get("general", "time_format")`. This is just a try but this is not working for `__repr__`. This is hard to explain ...

irispy/new_sji.py
BaptistePellorceAstro: ok I push it now

BaptistePellorceAstro:
I don't fix all I want, the is just for some reviews.

irispy/new_sji.py
BaptistePellorceAstro: I forget to change for `.get("X", None)` before the PR, but it's modified now and will be available for next PR : I don't want to spam PR for each modification.

irispy/new_sji.py
```diff
@@ -0,0 +1,203 @@
...
+               bandpass=self.meta["TWAVE1"],
+               date_start=self.meta["DATE_OBS"],
+               date_end=self.meta["DATE_END"],
+               instance_start=self.extra_coords["TIME"]["value"][0],
+               instance_end=self.extra_coords["TIME"]["value"][-1],
+               frame_num=self.meta["NBFRAMES"],
+               obs_id=self.meta["OBSID"],
+               obs_desc=self.meta["OBS_DESC"],
+               axis_types=self.world_axis_physical_types,
+               dimensions=self.dimensions, tmf=TIME_FORMAT))
```
BaptistePellorceAstro: Here is my problem.

I tried to modify one date to `None` to check how the code works. By changing the date to `None` is the same result of a missing value for `DATE_OBS` or `DATE_END` on the fits file. So, on the meta, I have `'DATE_OBS' = None` for example. If I ask SJI class to use `__repr__`, the time will be transferred to the `TIME_FORMAT = config.get("general", "time_format")`.

The problem is that `None` can't be transferred as a time format. The code will crash if one date is missing in the FITS file.

Hoping to be clearer !

BaptistePellorceAstro: Here is my problem.

I tried to modify one date to `None` to check how the code works. By changing the date to `None` is the same result of a missing value for `DATE_OBS` or `DATE_END` on the fits file. So, on the meta, I have `'DATE_OBS' = None` for example. If I ask SJI class to use `__repr__`, the time will be transferred to the `TIME_FORMAT = config.get("general", "time_format")`.

The problem is that `None` can't be transferred as a time format. The code will crash if one date is missing in the FITS file.

Hoping to be clearer !

irispy/new_sji.py
DanRyanIrish: Sure we can work through this.  First of all, the mask is not a nan mask.  The mask (```SJICube.mask```) is a boolean array (True/False)  with the same shape as the data array.  ```True``` means the corresponding pixel in the data array (```SJICube.data```) is masked, i.e. is bad data.  ```False``` means the opposite.  Using this boolean mask means whether a pixel is defined as good or bad is independent of its value in the data array.

The next issue is what value the unexposed pixels should have.  In truth, this doesn't actually matter because these pixels are masked.  All that matters is that we can identify them when we are creating the mask.  In the FITS files, the unexposed pixels have a value of ```-200``` (assuming data is scaled).  When I originally wrote the ``IRISSpectrograph``` class, I (perhaps unwisely) decided to change those values to NaN.  For the time being we can copy that convention in ```SJICube```.

So, finally to the specifics of the code.  Following the above convention, we want to set the data value of unexposed pixels to ```nan``` and set the mask values of those pixels to ```True```.  Previously this was simply done by finding which pixels have data values of ```-200```.  However, if memory mapping is turned on, unexposed pixels will no longer have a value of ```-200```, but some other value, let's call it ```BAD_PIXEL_VALUE_UNSCALED```.  So the change to the above code will be something like:
```
# Here I assume I've renamed BAD_PIXEL_VALUE to BAD_PIXEL_VALUE_SCALED at top of file.
if memmap is True:
    bad_pixel_value= BAD_PIXEL_VALUE_UNSCALED
elif memmap is False:
    bad_pixel_value = BAD_PIXEL_VALUE_SCALED
data_nan_masked[data == bad_pixel_value] = np.nan
mask = data_nan_masked == bad_pixel_value
```
Defining ```BAD_PIXEL_VALUE_UNSCALED``` must also be done.  But I've left other comments on this PR asking Tiago to confirm whether the scaling is always the same for SJI.  Could you ask him that question in person, as I'm not sure he gets these notifications?

irispy/new_sji.py
DanRyanIrish: ```.get("X", None)``` should be changed in this PR as it was originally to define the ```SJICube``` with a ```__repr__```.  Don't worry about spamming the PR.  That's how the system is set up :)

irispy/new_sji.py
BaptistePellorceAstro: Thanks, I have talked with him earlier and both of you helped me to have a better understanding of that. One problem on what you wrote is that if memmap is activated, we can't use `np.nan` "because it's not an integer". That why I modify it to 0 for now, hoping a better solving...

I go ask Tiago now.

irispy/new_sji.py
DanRyanIrish: Ah, yes I see.  I suggest in this case that you calculate the time values at the start of the start of the ```__repr__``` method, i.e. between current lines 94 and 95, and then put the result of that into the string.  This way you can use an if statement to handle cases where time is None.  So something like:
```
def __repr__(self):
    my_date = self.meta.get("MY_DATE", None)
    if my_date:
        my_string_date = # Convert to time string format here
    else:
        my_string_date = None
    return (
```
This way ```my_string_date``` is already converted to a time string (or is ```None```) and does not require formatting when generating the ```__repr__``` string itself.

irispy/new_sji.py
DanRyanIrish: That's a good point @BaptistePellorceAstro.  I suggest we don't in fact change the data values but only form the mask as part of this code snippet.  To be consistent, the same policy should be implemented when ```memmap is False``` and also in ```read_iris_spectrograph_level2_fits```.

irispy/new_sji.py
tiagopereira: I couldn't find an authoritative source on this, but for all the files I tested, the "masked" values in unscaled FITS units are always -32768. Makes sense since it is the lowest signed 16-bit integer. Also for all files (SJI and SG), `BSCALE` and `BZERO` seem to be the same, but  these we can always read from the header. I think it is safe to assume that the bad pixel value is always -32768.

irispy/new_sji.py
```diff
@@ -0,0 +1,209 @@
...
+from sunpy.time import parse_time
+from sunpy import config
+from irispy import iris_tools
+
+__all__ = ['SJICube']
+
+TIME_FORMAT = config.get("general", "time_format")
+
+# the following value is only appropriate for byte scaled images
+BAD_PIXEL_VALUE = -200
```
DanRyanIrish: I think to be clear, this variable should be renamed ```BAD_PIXEL_VALUE_SCALED```.

irispy/new_sji.py
DanRyanIrish: OK in that case the definition of ```BAD_PIXEL_VALUE_UNSCALED``` in this PR is fine :)

If we aren't comfortable just assuming what the ```BSCALE``` and ```BZERO``` we shall have to think how best to define the equivalency between the scaled SJI data unit, and the unscaled one.  I'll give this some thought.

irispy/new_sji.py
BaptistePellorceAstro: `.get("X", None)` added on new PR

irispy/new_sji.py
BaptistePellorceAstro: Is there a good function to do that ? I found some solutions but they are ugly ...

BaptistePellorceAstro: To convert the time as a good string format used by NDCube.

BaptistePellorceAstro: Solved

irispy/new_sji.py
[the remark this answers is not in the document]
DanRyanIrish: Use STARTOBS and ENDOBS as they denote the start and end of the Observing campaign, while DATE_OBS and END_OBS denote the start and end of the file.  In the case of the vast majority of SJI files, these two will be the same.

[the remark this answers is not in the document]
BaptistePellorceAstro: Ok, I am going to modify it now.

DanRyanIrish:
Now that we are including memory mapping in this PR, would it be useful to be able to relate the unscaled values to the scaled values?

If so, we should define an ```astropy``` unit equivalency between the scaled and unscaled SJI units.  The relationship is ```scaled = BSCALE * unscaled + BZERO```.  @tiagopereira has learned that it appears that ```BSCALE``` and ```BZERO``` are usually the same but that we should probably read these values from the FITS file where applicable.

One way forward is that we define a default SJI ```BSCALE``` and ```BZERO``` in ```iris_tools.py``` and then attach hidden bscale and bzero attributes of these values to the ```SJICube``` as part of the ```__init__```.  These would default to the default values in ```iris_tools```.  Then in the FITS reader function, the value can be taken from the FITS header and explicitly supplied to the ```__init__``` function.

Then we can define a hidden equivalency property to ```SJICube``` which defines the equivalency to the scaled unit.  Thus we will be able to use memory mapping but also have the ability to easily scale some values of the data.  Any thoughts?

DanRyanIrish:
@BaptistePellorceAstro if you feel the feature I mentioned above is a bit advanced, we can leave it for a future PR.  I don't think it should hold up this PR.  Let me know when you feel this current PR is ready for another code review.

BaptistePellorceAstro:
For now, I am writing a crop function inside the SJI class, I think to be close to solve the bug.

I don't start to watch on, I am focusing on solving crop bug. Maybe we can set standard values for BSCALE and BZERO to avoid links between iris_tools and fits files.

DanRyanIrish:
OK well we wont be able to merge that as it belongs in ```NDCube```.  But we can certainly copy it across to an ```ndcube``` if you have solved it the bug.

Yeah, I agree.  I think we can certainly have default values in ```iris_tools``` and then overwrite them in ```SJICube``` when an ```SJICube``` is created by ```read_iris_sji_level2_fits```.  That will keep ```iris_tools``` and ```SJICube``` independent of FITS files.

BaptistePellorceAstro:
Hi, this new version has the `crop_by_coords` function inside the SJICube class. 

I am waiting some comments about it because I have done a lot of work on and some things have to be modified to be understandable by everyone. I you have any question about anything, ask me and I will explain why I use that.

DanRyanIrish:
Hi @BaptistePellorceAstro.  Thanks for these latest additions.  It's great to see your Python improving!  However, I think the size of this PR is now starting to cause us problems.  We want to merge the solutions to the simpler things that are done/almost done (like the ```SJICube``` class with its ```__repr__```) before mixing them up with more difficult problems (like ```crop_by_coords```).  It was my understanding that you weren't going to add the ```crop_by_coords``` to this PR.  Clearly I misunderstood and that's probably my fault.  I think it would be good to take a step back now and reassess how to best get your contributions into IRISpy in the simplest and most efficient way.

I suggest we have another telecon early next week to do that.  I'll send an email around suggesting a time.

In the mean time, here are some things to consider that we can discuss at the telecon.  Firstly as I said, the size and multitude of things in this PR are making it cumbersome.  Where possible, PRs should fix a single bug/add a single feature or a single set of dependent bug fixes/features.  This way each dependent set of changes can be merged on their own merits without being blocked by issues with completely independent parts of the PR.  Of course this is the ideal.  There are many real cases where you don't fully achieve that, but that should always be the aim.  I think we should roll back this PR and get your work that is ready into the main repo as soon as possible.

Secondly, with regard to your ```crop_by_coords```, I think we should pause work on that for a number of reasons.  First, you've explicitly assumed 3 dimensions and the order of those dimensions.  This may be OK for slit-jaw data, but cannot be assumed in the general case handled by ```NDCube```.  Therefore, we can't copy the work across to ```NDCube```.  Secondly, when the bug in ```NDCube.crop_by_coords``` is fixed, the work you've done so far will be overwritten by that.  This seems like a waste of effort and I'm sure we all want the valuable hours you are putting into this project to be the most efficient and impactful.  The reason it's important to solve the general case is that if we had a separate implementation of ```crop_by_coords``` in ```SJICube```, and ```IRISSpectrogram``` and ```NDCube```, we've tripled the amount of code we need to maintain.  Over time, things will break and need to be fixed in three separate places.  This will cause IRISpy to be become a mess and perhaps even unusable.  So although it feels easier now to just make these assumptions for ```SJICube```, it will make it harder to maintain in the future and again, reduce the impact of the valuable hours your are putting into this project.

The last thing about your version of ```crop_by_coords``` is that it changes the API (Application Programming Interface -- basically the way to put in inputs) relative the the ```NDCube``` version.  That means that when ```NDCube.crop_by_coords``` is fixed and replaces the ```SJICube.crop_by_coords```, the API will break for people using it.  We should avoid such breakages as often as makes sense as people in the future may write tools that assume one API that they will then have to rewrite.  So we would be increasing the total amount of development effort to the community.  Sometimes API breakages are unavoidable, or worthwhile if they brings new power to the tools.  But where it's avoidable we should avoid it.

Issues like these are almost unavoidable as you learn the process of contributing to SunPy/SunPy-affiliated packages.  I have certainly made such mistakes several times.  You learn by experience.  So don't be disheartened.  Your contributions are very valuable.  We'll chat next week, reassess and then plan out the best way forward.

BaptistePellorceAstro:
Hi, except for few lines, I think this crop by coords can replace the existing one. I clearly understand what you mind about wasting time but I though that this work help me a lot to understand python and all the packages I used. How can I go back in this pull request ?

DanRyanIrish:
Yes, making mistakes and going down dead ends can be a valuable way to learn :)

We can discuss where to go back to on the telecon but if you want to do it sooner, I would suggest creating a new branch from your master branch that keeps the work you've done so far:
```git checkout -b sji_crop_coords```
This will also change you onto this new branch.  So next go back your original branch (in this case the master branch) by doing:
```git checkout master```
Then you can either roll back to the last commit commit before ```crop_by_coords``` was added by doing:
```git rest --hard <commit-has-number>```
where ```<commit-has-number>``` is the hexidecimal commit ID (CAUTION: This will permanently erase from the master branch any work you've done after that commit.  It will remain on the new branch you've created).  Or if it's easier you just delete the ```crop_by_coords``` method once you've switched back to the master branch.

**Comment by BaptistePellorceAstro** [comment]
Ok, thanks ! I will do it on Monday.

**Comment by BaptistePellorceAstro** [comment]
Hi, I removed `crop_by_coords` from the file. If you don't want some changes, I think this work can be merged.

In `__repr__`, about reading `instance_start` and `instance_end`, how I can set `None` if the `self.extra_coords["TIME"]["value"][0]` is not existing ? For now, an error can comes from here.

**Comment by DanRyanIrish** [comment]
Let's talk about that at our telecon in 15 minutes, @BaptistePellorceAstro.

**Comment by BaptistePellorceAstro** [comment]
Hi, You can find here the code I created on iris_tools.py

First I have create default values according to FITS files values (I don't know how you want to link them to readable values from FITS files)
`BSCALE = 0.25`
`BZERO = 7992`

Then, I follow the unit definition existing, and I create this new line for the new unit (Is it the good way to create this unit ?). This is inside `DN_UNIT` dictionary. 
`"SJI_UNSCALED": u.def_unit( "DN_IRIS_SJI_UNSCALED", (DETECTOR_GAIN["SJI"] / DETECTOR_YIELD["SJI"] * BSCALE + BZERO) * u.photon)}`

I think I have to modify the reading noise too, so I wrote that, is it correct ? (I just remove the existing code to be clearer but the code is not removed from the file)
`READOUT_NOISE = {"SJI_UNSCALED" : 1.2*DN_UNIT["SJI_UNSCALED"]}`

**Comment by DanRyanIrish** [comment]
Hi @BaptistePellorceAstro.  It looks like you still haven't pushed the changes.  Could you do that as well as the description you gave in the comments?  (Thanks for those, by the way!)

**Comment by DanRyanIrish** [comment]
Hi @BaptistePellorceAstro.  This PR looks great!  I've left a few comments asking for some final changes.  Once they're taken care of, I think we'll be pretty much ready to merge!

**Comment by BaptistePellorceAstro** [comment]
Ok, thanks ! I will modify that tomorrow at the morning. Sorry, I forgot to modify `new_sji.py` before push.

File: irispy/iris_tools.py [hunk 21]
```diff
@@ -24,12 +24,16 @@
 # paper.
 DETECTOR_GAIN = {"NUV": 18., "FUV": 6., "SJI": 18.}
 DETECTOR_YIELD = {"NUV": 1., "FUV": 1.5, "SJI": 1.}
+BSCALE = 0.25
+BZERO = 7992
```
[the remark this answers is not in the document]
BaptistePellorceAstro: Renamed

irispy/new_sji.py
```diff
@@ -0,0 +1,211 @@
...
+
+from datetime import timedelta
+
+import numpy as np
+from astropy.io import fits
+import astropy.units as u
+from astropy.wcs import WCS
+from sunpy.time import parse_time
+from irispy import iris_tools
+from ndcube import NDCube
```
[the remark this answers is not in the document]
BaptistePellorceAstro: Moved, now I understand why there is different blocks thanks !

irispy/new_sji.py
```diff
@@ -0,0 +1,211 @@
+'''
+This module provides movie tools for level 2 IRIS SJI fits file
+'''
+
+from datetime import timedelta
```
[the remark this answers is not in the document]
BaptistePellorceAstro: Added

irispy/new_sji.py
```diff
@@ -0,0 +1,211 @@
...
+    >>> from irispy.data import sample
+    >>> sji = read_iris_sji_level2_fits(sample.SJI_CUBE_1400)
+    """
+
+    def __init__(self, data, wcs, uncertainty=None, unit=None, meta=None,
+                 mask=None, extra_coords=None, copy=False, missing_axis=None,
+                 scaled=None):
+        """
+        Initialization of Slit Jaw Imager
+        """
```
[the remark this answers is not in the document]
BaptistePellorceAstro: Added

irispy/new_sji.py
```diff
@@ -0,0 +1,211 @@
...
+    # Open a fits file
+    my_file = fits.open(filename, memmap=memmap, do_not_scale_image_data=memmap)
+    # Derive WCS, data and mask for NDCube from fits file.
+    wcs = WCS(my_file[0].header)
+    data = my_file[0].data
+    data_nan_masked = my_file[0].data
+    if memmap:
+        data_nan_masked[data == BAD_PIXEL_VALUE_UNSCALED] = 0
+        mask = data_nan_masked == BAD_PIXEL_VALUE_UNSCALED
+        scaled = False
```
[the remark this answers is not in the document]
BaptistePellorceAstro: Added

irispy/new_sji.py
```diff
@@ -0,0 +1,211 @@
...
+    elif not memmap:
+        data_nan_masked[data == BAD_PIXEL_VALUE_SCALED] = np.nan
+        mask = data_nan_masked == BAD_PIXEL_VALUE_SCALED
+        scaled = True
```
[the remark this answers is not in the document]
BaptistePellorceAstro: Added

irispy/new_sji.py
```diff
@@ -0,0 +1,211 @@
...
+    # Derive exposure time from detector.
+    exposure_times = my_file[1].data[:, my_file[1].header["EXPTIMES"]]
+    # Derive unit and readout noise from detector
+    unit = iris_tools.DN_UNIT["SJI"]
```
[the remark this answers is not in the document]
BaptistePellorceAstro: And what's about the reading noise ? Do I have to add it too with the unit (I defined a reading noise for `SJI_UNSCALED` in `iris_tools.py`) ?

Line removed

irispy/iris_tools.py
```diff
@@ -24,12 +24,16 @@
...
 DN_UNIT = {
     "NUV": u.def_unit("DN_IRIS_NUV", DETECTOR_GAIN["NUV"]/DETECTOR_YIELD["NUV"]*u.photon),
     "FUV": u.def_unit("DN_IRIS_FUV", DETECTOR_GAIN["FUV"]/DETECTOR_YIELD["FUV"]*u.photon),
-    "SJI": u.def_unit("DN_IRIS_SJI", DETECTOR_GAIN["SJI"]/DETECTOR_YIELD["SJI"]*u.photon)}
+    "SJI": u.def_unit("DN_IRIS_SJI", DETECTOR_GAIN["SJI"]/DETECTOR_YIELD["SJI"]*u.photon),
+    "SJI_UNSCALED": u.def_unit("DN_IRIS_SJI_UNSCALED",
+                               (DETECTOR_GAIN["SJI"]/DETECTOR_YIELD["SJI"]*BSCALE + BZERO)*u.photon)}
```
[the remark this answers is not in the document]
BaptistePellorceAstro: Do you mean that I have to create a new function that convert SJI unit to SJI_UNSCALED unit and vice versa by using equivalences ?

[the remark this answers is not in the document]
BaptistePellorceAstro: I created the below tuple :
`SJI_SCALING = (DN_UNIT["SJI"], DN_UNIT["SJI_UNSCALED"], lambda x: 7992.0*x,` 
`lambda x: -31964.0*x)`

Then, when I do 
`iris_tools.DN_UNIT['SJI'].to(iris_tools.DN_UNIT['SJI_UNSCALED'], equivalencies=iris_tools.SJI_SCALING)`

I have a false value returning (same value for different calculations).
`0.0022509848058525606`

The values 7992.0 and -31964.0 are correct but I will replaced them by default values when I will solved my bug, they help me to work faster. 

Do you have an idea about it ?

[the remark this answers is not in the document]
DanRyanIrish: Firstly, I think ```SJI_SCALING``` has to be a list of tuples.  In this case it would be a list of one tuple.  Not sure if that makes a difference.

Next, isn't the function to go from unscaled to scaled ```x * SJI_DEFAULT_BSCALE + SJI_DEFAULT_BZERO```?  And then the reverse would be ```(x - SJI_DEFAULT_BZERO) / SJI_DEFAULT_BSCALE```?  Am I wrong on this?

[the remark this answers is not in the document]
BaptistePellorceAstro: Yes, I have these function but its not working ...

irispy/new_sji.py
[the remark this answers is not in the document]
BaptistePellorceAstro: Added

irispy/iris_tools.py
```diff
@@ -24,14 +24,20 @@
...
     "FUV": u.def_unit("DN_IRIS_FUV", DETECTOR_GAIN["FUV"]/DETECTOR_YIELD["FUV"]*u.photon),
     "SJI": u.def_unit("DN_IRIS_SJI", DETECTOR_GAIN["SJI"]/DETECTOR_YIELD["SJI"]*u.photon),
     "SJI_UNSCALED": u.def_unit("DN_IRIS_SJI_UNSCALED",
-                               (DETECTOR_GAIN["SJI"]/DETECTOR_YIELD["SJI"]*BSCALE + BZERO)*u.photon)}
+                               (DETECTOR_GAIN["SJI"]/DETECTOR_YIELD["SJI"]*SJI_DEFAULT_BSCALE
+                                + SJI_DEFAULT_BZERO)*u.photon)}
+SJI_SCALING = ((DN_UNIT["SJI"],
+                DN_UNIT["SJI_UNSCALED"],
+                lambda x: (x - SJI_DEFAULT_BZERO) / SJI_DEFAULT_BSCALE,
+                lambda x: x * SJI_DEFAULT_BSCALE + SJI_DEFAULT_BZERO))
```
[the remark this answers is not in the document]
BaptistePellorceAstro: I modified the unit to what you suggest but it is still not working ...

On the terminal, I wrote : 
`iris_tools.DN_UNIT['SJI'].to(iris_tools.DN_UNIT['SJI_UNSCALED'], equivalencies=iris_tools.SJI_SCALING)`

Is it the good way to use `equivalencies` ?

My return is always `- 0.000564334085778781`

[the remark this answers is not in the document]
DanRyanIrish: Hi @BaptistePellorceAstro.  I've figured out what the problem is.  In fact there are a two:
* Firstly, your definition of the ```DN_IRIS_SJI_UNSCALED``` unit in L34-35 is still incorrectly inverted.
* Secondly, ```DN_IRIS_SJI``` and ```DN_IRIS_SJI_UNSCALED``` are currently both defined as a fixed number of photons.  This means they are already compatible units and so the equivalency is not being applied as we want.  The fix to this is set ```DN_IRIS_SJI_UNSCALED``` to counts (```u.ct```) instead of photons (```u.photon```) in L35.
This should fix it.  Let me know if this works for you.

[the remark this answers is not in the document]
BaptistePellorceAstro: Hi, for the expression, I have changed it in my file that is not update on github. Then, I test with `u.ct` and I have `'DN_IRIS_SJI' and 'DN_IRIS_SJI_UNSCALED' are not convertible`. 

Tiago has found a way to avoid the problem : creating the unit as  
`DN_UNIT = {"SJI": u.def_unit("DN_IRIS_SJI", DETECTOR_GAIN/DETECTOR_YIELD*u.photon),
                     "SJI_UNSCALED": u.def_unit("DN_IRIS_SJI_UNSCALED")}`

Then the unit is just a name but the equivalencies are working.

Maybe, I must set the `DN_UNIT["SJI"]` for everyone and convert data with memmap with the equivalencies function.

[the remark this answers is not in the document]
tiagopereira: This pull request is getting really long and hard to follow. I have never used the `units` package, and so far fail to see the usefulness of writing an equivalency that will require a user to type an extremely long line of code. These units are meaningless in any case. I'd just set the units to `SJI_UNSCALED` or `SJI_SCALED` and be done with it. If people want `SJI_SCALED`, don't use `memmap`. This is what IDL does and I don't see why we should spend so long thinking about a feature that very few will ever use.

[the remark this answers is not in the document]
DanRyanIrish: I agree this PR has gotten far too big.  We should merge it as soon as possible.

Regarding the ``DN_UNIT```, I like @tiagopereira's solution.  Let's go with that and then get this PR merged.

I think we should keep the equivalency now that it has been written.  I foresee useful applications in the future.  Regarding units in general, they are a requirement of SunPy and AstroPy and so I think we stick to that convention to continue to leverage work from those packages.  As a user in my own work, I find units invaluable and I'm sure other users will find the same.  From what @BaptistePellorceAstro says, it sounds like he has made all the necessary changes regarding unit and equivalencies and just needs to push his latest changes before we can merge.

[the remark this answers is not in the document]
BaptistePellorceAstro: Hi, so I will delete the expression of the unscaled unit as Tiago said and let the unit just with the name. I let it without any value for now ?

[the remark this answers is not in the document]
DanRyanIrish: Yes.  Define the unscaled unit with just a name and leave the ```SJI_SCALING``` as is, so long as it works.  It should.  Push those changes and then we can get it merged.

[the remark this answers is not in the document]
BaptistePellorceAstro: Ok, merging will be available at the european morning tomorrow.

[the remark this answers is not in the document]
BaptistePellorceAstro: I can't set no value in the `SJI_UNSCALED` unit because `uncertainty` is using it. I so input `u.photon` as value to avoid an error.

BaptistePellorceAstro:
I set `u.photon` as value for the `SJI_UNSCALED` value to avoid an error.

If everything is OK for you, we can merge.

DanRyanIrish:
Hi @BaptistePellorceAstro.  Ah, I see the problem.  I suggest the following final quick changes to simplify things so we can get this merged today:
* Instead of removing ```u.photon``` from the definition of ```DN_IRIS_SJI_UNSCALED```, replace it with ```u.ct``` (counts).  This way the equivalency will still be valid.
* Instead of setting uncertainty to ```0```, set it to ```None```.  ```0``` is a number and so implies a value.  ```None``` means the uncertainty hasn't been determined, which is in fact the case.
* Also set ```mask = None``` if ```memmap is True```.
I'll make a few specific line comments as well related to these changes as well.  Assuming there are no errors caused by the changes so far, I think we should merge then.

irispy/iris_tools.py
```diff
@@ -24,26 +24,44 @@
...
-    "SJI": u.def_unit("DN_IRIS_SJI", DETECTOR_GAIN["SJI"]/DETECTOR_YIELD["SJI"]*u.photon)}
-READOUT_NOISE = {"NUV": 1.2*DN_UNIT["NUV"], "FUV": 3.1*DN_UNIT["FUV"],
-                 "SJI": 1.2*DN_UNIT["SJI"]}
+    "NUV": u.def_unit("DN_IRIS_NUV",
+                      DETECTOR_GAIN["NUV"] / DETECTOR_YIELD["NUV"]*u.photon),
+    "FUV": u.def_unit("DN_IRIS_FUV",
+                      DETECTOR_GAIN["FUV"]/DETECTOR_YIELD["FUV"]*u.photon),
+    "SJI": u.def_unit("DN_IRIS_SJI",
+                      DETECTOR_GAIN["SJI"]/DETECTOR_YIELD["SJI"]*u.photon),
+    "SJI_UNSCALED": u.def_unit("DN_IRIS_SJI_UNSCALED")}
```
DanRyanIrish: Should be ```u.def_unit("DN_IRIS_SJI_UNSCALED", u.ct)```

irispy/iris_tools.py
```diff
@@ -24,26 +24,44 @@
...
+# Define an equivalency between SJI and SJI_UNSCALED units
+SJI_SCALING = [(DN_UNIT["SJI"],
+                DN_UNIT["SJI_UNSCALED"],
+                lambda x: (x - SJI_DEFAULT_BZERO) / SJI_DEFAULT_BSCALE,
+                lambda x: x * SJI_DEFAULT_BSCALE + SJI_DEFAULT_BZERO)]
+
+READOUT_NOISE = {"NUV": 1.2*DN_UNIT["NUV"],
+                 "FUV": 3.1*DN_UNIT["FUV"],
+                 "SJI": 1.2*DN_UNIT["SJI"],
+                 "SJI_UNSCALED" : 1.2*DN_UNIT["SJI_UNSCALED"]}
```
DanRyanIrish: Remove this line defining ```SJI_UNSCALED``` readout noise.  It is not needed now we are not calculating the uncertainties.

irispy/iris_tools.py
```diff
@@ -96,7 +114,7 @@ def get_iris_response(pre_launch=False, response_file=None, response_version=Non
         if not(os.path.isfile(response_file)):
             raise KeyError("Not a valid file path")
 
-    
+
```
DanRyanIrish: Delete this empt line.  No need to two empty lines.

irispy/new_sji.py
```diff
@@ -0,0 +1,218 @@
...
+
+    # Open a fits file
+    my_file = fits.open(filename, memmap=memmap, do_not_scale_image_data=memmap)
+    # Derive WCS, data and mask for NDCube from fits file.
+    wcs = WCS(my_file[0].header)
+    data = my_file[0].data
+    data_nan_masked = my_file[0].data
+    if memmap:
+        data_nan_masked[data == BAD_PIXEL_VALUE_UNSCALED] = 0
+        mask = data_nan_masked == BAD_PIXEL_VALUE_UNSCALED
```
DanRyanIrish: Change to ```mask = None```

irispy/new_sji.py
```diff
@@ -0,0 +1,218 @@
...
+        scaled = False
+        # The three next lines are still in development, values are false
+        unit = iris_tools.DN_UNIT["SJI_UNSCALED"]
+        readout_noise = iris_tools.READOUT_NOISE["SJI_UNSCALED"]
+        uncertainty = 0
```
DanRyanIrish: Change to ```uncertainty = None```

irispy/iris_tools.py
BaptistePellorceAstro: Modified

irispy/iris_tools.py
BaptistePellorceAstro: Removed

irispy/iris_tools.py
BaptistePellorceAstro: I don't know why a empty line appears here !?

Deleted

irispy/new_sji.py
BaptistePellorceAstro: Changed

irispy/new_sji.py
BaptistePellorceAstro: Changed

BaptistePellorceAstro:
Changes has been made, ready to merge.

DanRyanIrish:
Terrific!  We are now running the automatic tests to make sure it hasn't broken anything that is already in the repo.  It shouldn't have since this PR creates a whole new module.  Assuming that's the case, we can merge once the tests pass.

DanRyanIrish:
So there are some errors but I think they are not actually because of your PR.

lgtm 👍

PR #70, opened 2018-04-19
fix the docs for sunpy 0.9

Cadair:


DanRyanIrish:
What is this doing and why is it required?

Cadair:
it updates irispy to reflect changes in the sphinx theme, so that the header bar of irispy matches all the other sunpy pages.

DanRyanIrish:
Could this be the cause of some of those weird errors we were discussing earlier?

Cadair:
well the docs build passed.

DanRyanIrish:
That was IRISpy.

PR #71, opened 2018-04-24
Adding a time correction function to SJICube

BaptistePellorceAstro:
Here the first version of the time correction function for SJICube.

### DIFF
--- a/irispy/new_sji.py
+++ b/irispy/new_sji.py
@@ -137,7 +137,7 @@ class SJICube(NDCube):
                axis_types=self.world_axis_physical_types,
                dimensions=self.dimensions))
 
-    def apply_exposure_time_correction(self, undo=False, copy=False, force=False):
+    def apply_exposure_time_correction(self, undo=False, force=False):
         """
         Applies or undoes exposure time correction to data and uncertainty and adjusts unit.
 
@@ -172,10 +172,13 @@ class SJICube(NDCube):
             applied (undone).
 
         """
+        # Raise an error if this method is called while memmap is used
+        if not self.scaled:
+            raise ValueError("This method is not available as you are using memmap")
         # Get exposure time in seconds and change array's shape so that
         # it can be broadcast with data and uncertainty arrays.
         exposure_time_s = u.Quantity(self.extra_coords["EXPOSURE TIME"]["value"], unit='s').value
-        if not isinstance(self.extra_coords["EXPOSURE TIME"]["value"], float):
+        if not np.isscalar(self.extra_coords["EXPOSURE TIME"]["value"]):
             if self.data.ndim == 1:
                 pass
             elif self.data.ndim == 2:
@@ -194,20 +197,11 @@ class SJICube(NDCube):
             new_data_arrays, new_unit = iris_tools.calculate_exposure_time_correction(
                 (self.data, self.uncertainty.array), self.unit, exposure_time_s, force=force)
         # Return new instance of SJICube with correction applied/undone.
-        if copy is True:
-            return SJICube(
-                data=new_data_arrays[0], wcs=self.wcs, uncertainty=new_data_arrays[1],
-                unit=new_unit, meta=self.meta, mask=self.mask, missing_axis=self.missing_axis,
-                extra_coords=convert_extra_coords_dict_to_input_format(self.extra_coords,
-                                                                       self.missing_axis))
-        else:
-            for i in range(self.data.shape[0]):
-                for j in range(self.data.shape[1]):
-                    for k in range(self.data.shape[2]):
-                        if not self.data[i, j, k] == np.nan:
-                            self.data[i, j, k] = new_data_arrays[0][i, j, k]
-                        if not self.uncertainty.array[i, j, k] == np.nan:
-                            self.uncertainty.array[i, j, k] = new_data_arrays[1][i, j, k]
+        return SJICube(
+            data=new_data_arrays[0], wcs=self.wcs, uncertainty=new_data_arrays[1],
+            unit=new_unit, meta=self.meta, mask=self.mask, missing_axis=self.missing_axis,
+            extra_coords=convert_extra_coords_dict_to_input_format(self.extra_coords,
+                                                                   self.missing_axis))
 
 
 def read_iris_sji_level2_fits(filename, memmap=False):
--- a/irispy/new_sji.py
+++ b/irispy/new_sji.py
@@ -11,6 +11,7 @@ import astropy.units as u
 from astropy.wcs import WCS
 from sunpy.time import parse_time
 from ndcube import NDCube
+from ndcube.utils.cube import convert_extra_coords_dict_to_input_format
 
 from irispy import iris_tools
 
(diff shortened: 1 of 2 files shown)


BaptistePellorceAstro:
For now, if `memmap = True` we can't use this function due to the use of `self.uncertainty`

irispy/new_sji.py
```diff
@@ -136,6 +137,78 @@ def __repr__(self):
...
+            If copy=False, the original SJICube is modified with the exposure
+            time correction applied (undone).
+            If copy=True, a new SJICube is returned with the correction
+            applied (undone).
+
+        """
+        # Get exposure time in seconds and change array's shape so that
+        # it can be broadcast with data and uncertainty arrays.
+        exposure_time_s = u.Quantity(self.extra_coords["EXPOSURE TIME"]["value"], unit='s').value
+        if not isinstance(self.extra_coords["EXPOSURE TIME"]["value"], float):
```
DanRyanIrish: Instead of ```isinstance(..., float)```,  this should be ```np.isscalar(...)```.  This way we don't get caught by edges cases where the value is ```int```, ```np.int64```, etc.

irispy/new_sji.py
```diff
@@ -136,6 +137,78 @@ def __repr__(self):
...
+                extra_coords=convert_extra_coords_dict_to_input_format(self.extra_coords,
+                                                                       self.missing_axis))
+        else:
+            for i in range(self.data.shape[0]):
+                for j in range(self.data.shape[1]):
+                    for k in range(self.data.shape[2]):
+                        if not self.data[i, j, k] == np.nan:
+                            self.data[i, j, k] = new_data_arrays[0][i, j, k]
+                        if not self.uncertainty.array[i, j, k] == np.nan:
+                            self.uncertainty.array[i, j, k] = new_data_arrays[1][i, j, k]
```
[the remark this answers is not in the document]
DanRyanIrish: Yes, I'm not surprised this is taking forever with the number of nested for loops you have.  Regarding the ```Can't set attribute``` error, I believe the ```.data``` attribute, which we inherit from ```astropy.nddata.NDData``` can be operated on but not reset.  Therefore, to simplify things, I would delete this else statement altogether so the function always reinstantiates the ```SJICube``` object as you do in the ```if``` part of the statement.  (The ```if``` line (L197) and the ```copy``` kwarg would have to be deleted.)  Then if the user wants to replace the current instance, they can simply do
```my_cube = my_cube.apply_exposure_time_correction()```
instead of
```my_cube.apply_exposure_time_correction()```
This would make it consistent with ```IRISSpectrogram```.  It's also the standard behaviour exhibited by sunpy objects, i.e. if they have a method that operates on their data, that method returns a new copy of the object with the operation performed, rather than overwriting the existing object.  If there's a good case to do something different we can consider it.  But the "return new copy" behaviour should be the default or only option.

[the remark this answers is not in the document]
Comment (DanRyanIrish) [reply irispy/new_sji.py]: Alternatively, there is in fact a simple solution if you do want to keep this copy option.  That is:
```python
if copy:
    pass
else:
    self.data[:] = new_data_arrays[0][:]
    self.uncertainty[:] = new_data_arrays[1][:]

**Comment by DanRyanIrish** [comment]
@BaptistePellorceAstro, regarding your comment about when ```memmap is True```, that's right.  But I believe it's also the case that by using this method, we will end up reading all the data into RAM anyway, as the data must be in RAM to operate on it, e.g., divide by the exposure.

So we should decide whether we should raise an error if this method is called and ```memmap is True```.

File: irispy/new_sji.py (hunk 2 above)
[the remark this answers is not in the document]
Comment (DanRyanIrish) [reply irispy/new_sji.py]: Also, if we do keep this copy behaviour it should also be extended to ```IRISSpectrogram``` so it and ```SJICube``` behave consistently.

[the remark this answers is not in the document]
Comment (BaptistePellorceAstro) [reply irispy/new_sji.py]: Ok, thanks. I will check all of that tomorrow.

**Comment by BaptistePellorceAstro** [comment]
Nice, I will look about all you said. Thanks !

File: irispy/new_sji.py (hunk 1 above)

**Comment by BaptistePellorceAstro** [comment]
What kind of test do you want I create for the new method ?

**Comment by DanRyanIrish** [comment]
Similarly to your ndcube PR, we want a test simple test takes input and expected values, executes the method using the input, and compares the output to the expected values.  Then we want to use the ```parameterize``` decorator to put in different combinations of input and expected values to test the different API or input options.  This will required that different "fake" ```SJICube```s are created at the top of the file with different features.  Look at the code and work out what type of input would be needed for the code to enter all the different if statements, etc.

Then we will also want a simple test that checks the correct errors are raised.  Again this will need different ```SJICube```s as inputs.

**Comment by BaptistePellorceAstro** [comment]
I improved the return of the method to avoid a too long calculation time. Then I created a new test file for `new_sji.py`. I copied some test from the other test file for `sji.py` but I don't know which test add in this method. Waiting your returns.

**Comment by DanRyanIrish** [comment]
The new test file should currently only have two tests in it: ```test_apply_exposure_time_correction``` and ```test_apply_exposure_time_correction_error``` which test the correct answers and errors are returned by ```SJICube.apply_exposure_time_correction```.

**Comment by BaptistePellorceAstro** [comment]
Ok, I am working on it.

**Comment by DanRyanIrish** [comment]
An additional point.  You can see, line by line, what parts of code are tested by your tests.  In the top level ndcube directory (where the setup.py file is) run the following in the command line (not in a python session):
```python setup.py test --coverage```
This runs all the tests and creates a coverage report.  To see the report, go to ```ndcube/htmlcov/``` and open the index.html file in a browser.  This gives you a list of all the non-test files in the repo and gives the percentage of lines covered by testing.

In this case, to see the effect of you exposure time tests, click on the new_sji.py and it brings up an html version of the code.  Scroll to the ```SJICube.apply_exposure_time``` method.  Highlighted red lines are lines that the tests don't ever execute.  You can use this to determine what other tests you need to write.

(N.B. There will be a lot of red highlighted lines outside the ```SJICube.apply_exposure_time``` as no other tests have been written.  Don't worry about writing tests for those parts of the code for the time being.)

A word of warning, this method doesn't always tell you every single test you want to implement.  Occasionally there might be different cases the user could input that might not cause any new lines of code to be tested, but still could cause an error.  You would still want to test these even if it doesn't increase the coverage.  Having said this, the coverage report is still a very useful tool for building your tests.

You can do the same thing in the ndcube repo too.

**Comment by BaptistePellorceAstro** [comment]
Nice, I continue to work on my test due to an error then I will check what you said. Thanks !

File: irispy/tests/test_new_sji.py [hunk 3]
```diff
@@ -0,0 +1,105 @@
...
+import numpy as np
+from ndcube.utils.wcs import WCS
+
+from irispy import iris_tools
+from irispy.new_sji import SJICube, read_iris_sji_level2_fits
+
+# Sample data for tests
+private_fits_path = '/mn/stornext/u3/baptistp/iris/level2/2014/12/11/20141211_191222_3803105278/'
+file_name = 'iris_l2_20141211_191222_3803105278_SJI_1330_t000.fits'
+file_path = private_fits_path + file_name
```
DanRyanIrish: This raises an important point that we should discuss, i.e. whether we should store files of test data in the repo. @Cadair has had this discussion a few times in SunPy.  Any thoughts @Cadair?

DanRyanIrish:
@BaptistePellorceAstro Aside from the question of the test data file, let us know when the you feel this is ready for another proper review.

DanRyanIrish:
I think it's looking close!

irispy/tests/test_new_sji.py
DanRyanIrish: I think in order to get this PR merged, I'd suggest removing this definition of the FITS file and also removing the ```test_read_iris_sji_level2_fits```.  Then you can submit a new PR with ```test_read_iris_sji_level2_fits``` and we can deal with the issue of test data files separately.

irispy/tests/test_new_sji.py
```diff
@@ -0,0 +1,105 @@
...
+cube_F = SJICube(data, wcs, uncertainty=uncertainty, mask=mask_cube, unit=unit,
+                 extra_coords=extra_coords, scaled=scaled_F)
+cube_4D = SJICube(data_4D, wcs_4D, uncertainty=uncertainty, mask=mask_4D, unit=unit,
+                  extra_coords=extra_coords, scaled=scaled_T)
+
+# Tests
+@pytest.mark.parametrize("sji_file", [
+    (file_path, False),
+    (file_path, True)])
+def test_read_iris_sji_level2_fits(sji_file):
```
DanRyanIrish: With reference to the above comment, remove this test from this PR and submit it in a new PR so we don't get bogged down on how to handle test data files.

DanRyanIrish:
I'm happy to merge this :)  Any final comments @Cadair?

DanRyanIrish:
Congratulation @BaptistePellorceAstro on another merged PR!

PR #74, opened 2018-04-27
Creation of dust mask function

BaptistePellorceAstro:
Hi @DanRyanIrish , some lines to begin to speak about the dust mask function. For now, I set `NaN` to values under 0.5 that is match with dust particles position. This will be modified, that was just to check some code lines.

What kind of dust mask do you want ? A mask ? `NaN` data ? A file ?

### DIFF
--- a/irispy/iris_tools.py
+++ b/irispy/iris_tools.py
@@ -829,7 +829,6 @@ def calculate_dust_mask(data_array, mask=None):
     # Set the pixel value to True is the pixel is recognized as a dust pixel.
     mask[(data_array < 0.5) & (data_array > -200)] = True
     # Extending the mask to avoid the neighbours pixel influenced by the dust pixels.
-    struct = ndimage.generate_binary_structure(2, 2)
-    for i in range(data_array.shape[0]):
-        mask[i] = ndimage.binary_dilation(mask[i], structure=struct).astype(mask.dtype)
+    struct = np.array([np.zeros((3, 3)), np.ones((3, 3)), np.zeros((3, 3))], dtype=bool)
+    mask = ndimage.binary_dilation(mask, structure=struct).astype(mask.dtype)
     return mask
--- a/irispy/iris_tools.py
+++ b/irispy/iris_tools.py
@@ -804,7 +804,7 @@ def _apply_or_undo_exposure_time_correction(sequence, correction_function):
             converted_data_list.append(cube)
     return converted_data_list
 
-def calculate_dust_mask(data_array, mask=None):
+def calculate_dust_mask(data_array, input_mask=None):
     """Calculate a mask with the dust positions in a given arrayself.
 
     Parameters
@@ -820,15 +820,16 @@ def calculate_dust_mask(data_array, mask=None):
         when the value is True.
 
     """
-    # Check or set up of the mask array with the same shape than the data array.
-    if mask:
-        if mask.shape != data_array.shape:
-            raise ValueError("The data array and the mask array must have the same shape")
-    else:
-        mask = np.zeros_like(data_array, dtype=bool)
+    # Creating a mask with the same shape than the inputed data array.
+    mask = np.zeros_like(data_array, dtype=bool)
     # Set the pixel value to True is the pixel is recognized as a dust pixel.
     mask[(data_array < 0.5) & (data_array > -200)] = True
     # Extending the mask to avoid the neighbours pixel influenced by the dust pixels.
     struct = np.array([np.zeros((3, 3)), np.ones((3, 3)), np.zeros((3, 3))], dtype=bool)
     mask = ndimage.binary_dilation(mask, structure=struct).astype(mask.dtype)
+    # Check if the inputed mask array has the same shape than the inputed data array.
+    if isinstance(input_mask, np.ndarray):
+        if input_mask.shape != data_array.shape:
+            raise ValueError("The data array and the mask array must have the same shape")
+        mask[input_mask] = True
     return mask
--- a/irispy/iris_tools.py
+++ b/irispy/iris_tools.py
@@ -804,7 +804,7 @@ def _apply_or_undo_exposure_time_correction(sequence, correction_function):
             converted_data_list.append(cube)
     return converted_data_list
 
-def calculate_dust_mask(data_array, input_mask=None):
+def calculate_dust_mask(data_array):
     """Calculate a mask with the dust positions in a given arrayself.
 
     Parameters
@@ -827,9 +827,4 @@ def calculate_dust_mask(data_array, input_mask=None):
     # Extending the mask to avoid the neighbours pixel influenced by the dust pixels.
     struct = np.array([np.zeros((3, 3)), np.ones((3, 3)), np.zeros((3, 3))], dtype=bool)
     mask = ndimage.binary_dilation(mask, structure=struct).astype(mask.dtype)
-    # Check if the inputed mask array has the same shape than the inputed data array.
-    if isinstance(input_mask, np.ndarray):
-        if input_mask.shape != data_array.shape:
-            raise ValueError("The data array and the mask array must have the same shape")
-        mask[input_mask] = True
     return mask
--- a/irispy/iris_tools.py
+++ b/irispy/iris_tools.py
@@ -14,6 +14,7 @@ from astropy.modeling import fitting
 from astropy.modeling.models import custom_model
 from astropy import constants
 import scipy.io
+from scipy import ndimage
 from scipy import interpolate
 from sunpy.time import parse_time
 import sunpy.util.config
@@ -802,3 +803,25 @@ def _apply_or_undo_exposure_time_correction(sequence, correction_function):
         else:
             converted_data_list.append(cube)
     return converted_data_list
+
+def calculate_dust_mask(data_array):
+    """Calculate a mask with the dust positions in a given arrayself.
+
+    Parameters
+    ----------
+    data_array : `numpy.ndarray`
+        This array contains some dust poisition that will be calculated. The array
+        must have scaled values.
+
+    Returns
+    -------
+    dust : `numpy.ndarray` of `bool`
+        This array has the same shape than data_array and contains the dust positions
+        when the value is True.
+
+    """
+    dust = (data_array < 0.5) & (data_array > -200)
+    struct = ndimage.generate_binary_structure(2, 2)
+    for i in range(data_array.shape[0]):
+        dust[i] = ndimage.binary_dilation(dust[i], structure=struct).astype(dust.dtype)
+    return dust
--- a/irispy/iris_tools.py
+++ b/irispy/iris_tools.py
@@ -804,7 +804,7 @@ def _apply_or_undo_exposure_time_correction(sequence, correction_function):
             converted_data_list.append(cube)
     return converted_data_list
 
-def calculate_dust_mask(data_array):
+def calculate_dust_mask(data_array, mask=None):
     """Calculate a mask with the dust positions in a given arrayself.
 
     Parameters
@@ -820,8 +820,16 @@ def calculate_dust_mask(data_array):
         when the value is True.
 
     """
-    dust = (data_array < 0.5) & (data_array > -200)
+    # Check or set up of the mask array with the same shape than the data array.
+    if mask:
+        if mask.shape != data_array.shape:
+            raise ValueError("The data array and the mask array must have the same shape")
+    else:
+        mask = np.zeros_like(data_array, dtype=bool)
+    # Set the pixel value to True is the pixel is recognized as a dust pixel.
+    mask[(data_array < 0.5) & (data_array > -200)] = True
+    # Extending the mask to avoid the neighbours pixel influenced by the dust pixels.
     struct = ndimage.generate_binary_structure(2, 2)
     for i in range(data_array.shape[0]):
-        dust[i] = ndimage.binary_dilation(dust[i], structure=struct).astype(dust.dtype)
-    return dust
+        mask[i] = ndimage.binary_dilation(mask[i], structure=struct).astype(mask.dtype)
+    return mask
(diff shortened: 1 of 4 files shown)


DanRyanIrish:
Hi @BaptistePellorceAstro.  Here are my thoughts on the dust mask method:
* The method should apply an algorithm to the data to locate the dust pixels, but not change the data values.  Instead it should change the boolean mask that SJICube already has.
* Perhaps, there could also be a kwarg that allows you mask or unmask those pixels, so the process is reversible.
* Perhaps there should also be an attribute added to ```SJICube```, like ```dust_masked``` which is either ```True``` or ```False``` and altered by the method.  This way users can quickly tell whether the dust mask is on or off.

Regarding the algorithm that should be used to find those dust pixels, I will refer to @tiagopereira.  I know there's such an algorithm in the SSW IDL IRIS software, developed by the IRIS team, but have never looked at it.  But I think we should mimic the mathematics of whatever the IRIS team has done.

@tiagopereira could elaborate on this here so we can better understand how this should be done?

BaptistePellorceAstro:
Yes thanks, that is clearer now !

DanRyanIrish:
Hi @BaptistePellorceAstro.  How are things coming along with this PR?  Do you need any guidance?

BaptistePellorceAstro:
Hi, I am finding an other way that `dust = self.data < 0.5` to check dust particles. The problem with that is some cosmic rays are taken as dust particles. So I try to find a way to take only the dust particles.

Moreover, I don't know IDL and we tried with Tiago to understand the codes but we are lost on `*.genx` files that we can't open.

BaptistePellorceAstro:
My problem is, for some frames, more of the half of the data are masked due to cosmic rays which are saturating the detector (they have a negative value). I tried with `ndimage.binary_erosion` but half of the dust particles are not detected. Do you have an idea about it ?

For now, I am masking the wrong data. Why not keep that ?

tiagopereira:
I found out how the bad pixel mask is read in IDL and how to do it in Python.

This is how it is read in IDL:
1. Load latest `*flat.genx` file with flat data. This file is written sequentially,   so the latest file has all the previous history.
2. Load latest `*badpix.geny` file with bad pixel mask. This file is also written sequentially, but before 20160825_031012 it was in genx format.
3. Find RECNUM in flat index corresponding to detector and closest time to desired IRIS file. 
4. Get bad pixel mask from `badpix.geny` given the RECNUM file.

An example how to read it in Python:

```
import numpy as np
from sunpy.io.special.genx import read_genx
from scipy.io import readsav

badpix_data = readsav("20180315_004203_badpix.geny")
flat_data = read_genx("20180120_012828_flat.genx")

# This happens to be the record number for SJI_2796 on 20180110
recnum = flat_data['SAVEGEN0'][-2]['RECNUM']

# This is a 1D array with the indices of bad pixels (flattened)
tmp = badpix_data.p0[0]['F' + str(recnum)]

# Put as a 2D boolean mask, pixel size should not change
badpix_mask = np.zeros((1096, 2072), dtype='bool')
badpix_mask.flat[tmp] = True  
```

This `badpix_mask` should be the final boolean mask.

In the example above I just chose a given record number, but a reader function should match `recnum` to the required detector and time. Probably a good idea to convert `flat_data` from an object array of ordered dicts to a pandas `DataFrame`.

A few notes:
- `badpix_mask` is for the full detector (e.g. for SJIs on the left side is used)
- `badpix_mask` is also not binned, will need to do this manually if needed
- `badpix_mask` is really for level 1 files. With SJI, the window of level 2 files will be moving around in cases of rasters, so `badpix_mask` will also have to be moved.

DanRyanIrish:
Thanks for looking into this @tiagopereira.  I'm sure this will be very helpful!

BaptistePellorceAstro:
I spoke with Tiago about the dust_mask and we think that we waste time by trying the last method and we should use my version (with negative values). What do you think about that ? 
I will try to select the cosmic rays which have positive values.

DanRyanIrish:
Why do you think this method is a waste of time?  What are the issues?

BaptistePellorceAstro:
I know this is very heavy to run but the result looks good so if you have some ideas to enhance that, I will be happy. For now, I am looking about some filters, maybe they can help me.

BaptistePellorceAstro:
About what Tiago sent before, the mask is nice but here the issues :
- The mask is shifted : I succeeded to align the dust positions but I am afraid that the values I used are not great in a general case.
- Some of our data are rotated : I succeeded to rotate the mask but some bad pixels can be found.
- Shape : Some of our data don't have the same shape that the mask shape (500x500 and 1100x1100 for example)
- Field of view : For some observation, the field of view is moving and I don't mind a way to solve that.

To conclude, I think this way is requiring a lot of work that I can easily do whit the code I commits today. What is your opinion on that ?

DanRyanIrish:
I outlined a way forward on the dust mask in Issue #18.  Let me know if you have any questions or comments.

BaptistePellorceAstro:
I created a new method in `iris_tools` which is `calculate_dust_mask`. This method take one array and return a boolean array, with the same shape, where True is a dust position.

Then, I called this method in `IRISMapCube.__init__` and I save the return in `self.dust_mask` as we just have to calculate one time the mask.

I created the `apply_dust_mask` method in `IRISMapCube` to apply/remove the dust positions inside `self.mask`.

This mask contains ONLY the dust positions, not the cosmic rays, and run only for scaled values.

irispy/iris_tools.py
```diff
@@ -802,3 +803,25 @@ def _apply_or_undo_exposure_time_correction(sequence, correction_function):
         else:
             converted_data_list.append(cube)
     return converted_data_list
+
+def calculate_dust_mask(data_array):
```
DanRyanIrish: I think this should have the option to supply a mask array and then return an altered copy of that mask with the dust pixels marked.  If a mask array isn't supplied, one can be generated.  So in practice this would be like:
``` python
def calculate_dust_mask(data_array, mask_array=None):
    if mask_array is None:
        mask_array = np.zeros_like(data_array, dtype=bool)
```

irispy/iris_tools.py
```diff
@@ -802,3 +803,25 @@ def _apply_or_undo_exposure_time_correction(sequence, correction_function):
...
+    -------
+    dust : `numpy.ndarray` of `bool`
+        This array has the same shape than data_array and contains the dust positions
+        when the value is True.
+
+    """
+    dust = (data_array < 0.5) & (data_array > -200)
+    struct = ndimage.generate_binary_structure(2, 2)
+    for i in range(data_array.shape[0]):
+        dust[i] = ndimage.binary_dilation(dust[i], structure=struct).astype(dust.dtype)
```
DanRyanIrish: I think this can be simplified to the following:
```python
mask_array[(data_array < 0.5) & (data_array > -200)] = True
return mask
```

DanRyanIrish:
Hi @BaptistePellorceAstro.  After more thought, I think this implementation shouldn't rely on a fixed dust mask array.  Otherwise the dust mask would have to be folded into the slicing of the object which is a lot more work.  It is also more RAM required, especially if the data are large.

Instead, I think the any time the ```self.dust_mask``` is used, it should be replaced with a call to ```calculate_dust_mask```.  If the same dust mask is used multiple times in a function, it can be set to a variable.  But it shouldn't be attached to the object as ```self.dust_mask```.

What are your thoughts?

irispy/iris_tools.py
BaptistePellorceAstro: I am not agree because a dust pixel can affect his neighbours so I have to use `dilate` to remove these pixels.

BaptistePellorceAstro:
This is a good idea but I think I will create a new variable to store the `dust_mask` to avoid to call `calculate_dust_mask` more than one time.

About your suggestion to calculate the mask, I am not agree because a dust pixel can affect his neighbours so I have to use `dilate` to remove these pixels.

irispy/iris_tools.py
BaptistePellorceAstro: You want to do that if the user just want the dust mask and not the previous mask ?

irispy/iris_tools.py
BaptistePellorceAstro: The problem of your code line is that mask_array is now a 1D array of boolean, that is why I used something else.

tiagopereira: You should not need a loop here. You can use `binary_dilation` with any array shape, just make sure that `struct` is appropriate (e.g. zero along the dimensions you don't want to dilate).

DanRyanIrish: Ah OK.  I didn't get the significance of the ```dilate```.  However, just for reference, my example shouldn't result in a 1D array.  The shape should be maintained.  That could be useful for you in the future.

DanRyanIrish:
Hi @BaptistePellorceAstro, the problem having ```self.dust_mask``` is what happens when you slice the ```IRISMapCube```?  Since the ```self.dust_mask``` attribute isn't handled in the ```__getitem___```, ```apply_dust_mask``` will break if you run it after slicing.  So to stop that happening, you wont be able to rely on ```self.dust_mask``` and instead will have to call the dust mask function.

irispy/iris_tools.py
BaptistePellorceAstro: Oh yes, I don't know why I think to that thanks Tiago.

BaptistePellorceAstro:
Oh yes, I understand now the problem of `self.dust_mask`. I will remove it.

irispy/iris_tools.py
BaptistePellorceAstro: Yes Dan, I was not using the first lines you wrote so that was not working. I modified it so.

BaptistePellorceAstro:
If I well understand, this is something like that you want ?

DanRyanIrish:
This is looking good @BaptistePellorceAstro.  Now it needs tests.

BaptistePellorceAstro:
Nice ! I go mind about that now.

BaptistePellorceAstro:
Here the tests for `apply_dust_mask`. 

I think, for a better visibility, that we should separate the data sample in the test file by tests as I did here. When I started to work on tests I was completely lost in the sample. I think this should help beginners in this file. What you think about that ?

irispy/new_sji.py
```diff
@@ -200,6 +203,29 @@ def apply_exposure_time_correction(self, undo=False, force=False):
...
+        undo: `bool`
+            If False, dust particles positions mask will be applied.
+            If True, dust particles positions mask will be removed.
+            Default=False
+
+        Returns
+        -------
+        result :
+            Rewrite self.mask with/without the dust positions.
+        """
```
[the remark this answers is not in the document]
BaptistePellorceAstro: Rewrited

irispy/iris_tools.py
```diff
@@ -802,3 +803,33 @@ def _apply_or_undo_exposure_time_correction(sequence, correction_function):
...
+    # Set the pixel value to True is the pixel is recognized as a dust pixel.
+    mask[(data_array < 0.5) & (data_array > -200)] = True
+    # Extending the mask to avoid the neighbours pixel influenced by the dust pixels.
+    struct = np.array([np.zeros((3, 3)), np.ones((3, 3)), np.zeros((3, 3))], dtype=bool)
+    mask = ndimage.binary_dilation(mask, structure=struct).astype(mask.dtype)
+    # Check if the inputed mask array has the same shape than the inputed data array.
+    if isinstance(input_mask, np.ndarray):
+        if input_mask.shape != data_array.shape:
+            raise ValueError("The data array and the mask array must have the same shape")
+        mask[input_mask] = True
```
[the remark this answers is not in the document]
BaptistePellorceAstro: Removed

irispy/iris_tools.py
```diff
@@ -802,3 +803,33 @@ def _apply_or_undo_exposure_time_correction(sequence, correction_function):
         else:
             converted_data_list.append(cube)
     return converted_data_list
+
+def calculate_dust_mask(data_array, input_mask=None):
```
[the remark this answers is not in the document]
BaptistePellorceAstro: Removed

irispy/iris_tools.py
```diff
@@ -802,3 +803,33 @@ def _apply_or_undo_exposure_time_correction(sequence, correction_function):
...
+        This array has the same shape than data_array and contains the dust positions
+        when the value is True.
+
+    """
+    # Creating a mask with the same shape than the inputed data array.
+    mask = np.zeros_like(data_array, dtype=bool)
+    # Set the pixel value to True is the pixel is recognized as a dust pixel.
+    mask[(data_array < 0.5) & (data_array > -200)] = True
+    # Extending the mask to avoid the neighbours pixel influenced by the dust pixels.
+    struct = np.array([np.zeros((3, 3)), np.ones((3, 3)), np.zeros((3, 3))], dtype=bool)
```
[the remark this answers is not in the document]
BaptistePellorceAstro: To use `ndimage.binary_dilation`, I need to input a `structure` kwarg. This kwarg is like a template for the dilation. Here, our dust pixel is the `struct[1,1,1]` pixel. By setting `np.ones((3,3))` for `struct[1]`, we say that all neighbours (diagonal included) of our dust pixel must be considerate as a dust pixel. About `stuct[0]` and `struct[2]`, I added them as Tiago suggested me to avoid the previous loop I was using. Their translation is " In the previous and the next frame do : Nothing" due to the `np.zeros((3,3)).

The `structure` kwarg is only dependent on what we need to do as I said before. The most important thing is if we not loop, we must have a 3D kwarg.

Is it clearer now ?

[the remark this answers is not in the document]
DanRyanIrish: OK, this makes more sense now :)

BaptistePellorceAstro:
The test coverage is perfect, if checks are passed, I am ready to merge this PR after all !

DanRyanIrish:
The latest master version needs to be merged into this PR.  Then it'll be ready to merge if tests pass :)

Issue #79, opened 2018-05-02
Making repr consistent across objects, better repr for spectrograph


tiagopereira:
There is currently a lot of variation on what the `repr` of different IRISpy objects shows. I suggest this is made more consistent. For example, the `new_sji.SJICube` seems to be indented, and has redundant keywords such as "Observatory" (always IRIS) and "Instrument" (always SJI). The data dimensions also are shown in different ways, with the one in `IRISSpectrograph` being the least readable.

In fact, I think there is much room for improvement in the `repr` of `IRISSpectrograph`. A particular offender is the "Instance period", which is shown as a collection of datetime objects when it should be only the range, and formatted in ISO 8601 or similar.

For example, a `repr` of a spectrograph object I just tested opens up as the following:

```
<iris.IRISSpectrograph instance
OBS ID: 3882010194
OBS Description: Large coarse 64-step raster 126x120 64s  C II   Si IV   Mg II h/k
OBS period: 2018-02-04 01:27:45.420000 -- 2018-02-04 01:45:50.272000
Instance period: [datetime.datetime(2018, 2, 4, 1, 27, 45, 590000)
 datetime.datetime(2018, 2, 4, 1, 28, 2, 560000)
 datetime.datetime(2018, 2, 4, 1, 28, 19, 360000)
 datetime.datetime(2018, 2, 4, 1, 28, 36, 430000)
 datetime.datetime(2018, 2, 4, 1, 28, 53, 290000)
 datetime.datetime(2018, 2, 4, 1, 29, 10, 310000)
 datetime.datetime(2018, 2, 4, 1, 29, 27, 130000)
 datetime.datetime(2018, 2, 4, 1, 29, 44, 20000)
 datetime.datetime(2018, 2, 4, 1, 30, 0, 960000)
 datetime.datetime(2018, 2, 4, 1, 30, 18, 60000)
 datetime.datetime(2018, 2, 4, 1, 30, 34, 850000)
 datetime.datetime(2018, 2, 4, 1, 30, 51, 780000)
 datetime.datetime(2018, 2, 4, 1, 31, 8, 710000)
 datetime.datetime(2018, 2, 4, 1, 31, 25, 840000)
 datetime.datetime(2018, 2, 4, 1, 31, 42, 630000)
 datetime.datetime(2018, 2, 4, 1, 31, 59, 520000)
 datetime.datetime(2018, 2, 4, 1, 32, 16, 810000)
[...]

PR #81, opened 2018-05-04
Getting a better __repr__ for IRISSpectrograph

tiagopereira:
This is an attempt to fix some of the problems in #79.

### DIFF
--- a/irispy/spectrograph.py
+++ b/irispy/spectrograph.py
@@ -54,8 +54,8 @@ class IRISSpectrograph(object):
     def __repr__(self):
         spectral_window = self.spectral_windows["spectral window"][0]
         spectral_windows_info = "".join(
-            ["\n{0:>15} : {1:16} pix".format(name,
-                str([int(dim.value) for dim in self.data[name].dimensions[1::]]))
+            ["\n{0:>15} : {1:20} pix".format(name,
+                str([int(dim.value) for dim in self.data[name].dimensions]))
                 for name in self.spectral_windows["spectral window"]])
         obs_start = self.meta["STARTOBS"]
         obs_end = self.meta["ENDOBS"]
@@ -73,11 +73,11 @@ class IRISSpectrograph(object):
         obs_period, instance_period = result
         return ("<iris.IRISSpectrograph instance\nOBS ID: {obsid}\n"
                "OBS Description: {obsdesc}\n"
-               "OBS period: {obs_period}\n"
+               "OBS Period: {obs_period}\n"
                "Instance period: {inst_period}\n"
-               "Number unique raster positions: {nraster}\n"
-               "Number of repeats: {repeats}\n"
-               "Spectral windows : dimensions [raster axis, slit axis, spectral axis]:"
+               "OBS Number unique raster positions: {nraster}\n"
+               "OBS Number of repeats: {repeats}\n"
+               "Spectral windows : dimensions [repeats axis, raster axis, slit axis, spectral axis]:"
                "{spec}>").format(obsid=self.meta["OBSID"], obsdesc=self.meta["OBS_DESC"],
                    obs_period=obs_period,inst_period=instance_period,
                    repeats=int(self.data[spectral_window].dimensions[0].value),
--- a/irispy/spectrograph.py
+++ b/irispy/spectrograph.py
@@ -54,18 +54,27 @@ class IRISSpectrograph(object):
     def __repr__(self):
         spectral_window = self.spectral_windows["spectral window"][0]
         spectral_windows_info = "".join(
-            ["\n    {0}\n        (raster axis, slit axis, spectral axis) {1}".format(
-                name,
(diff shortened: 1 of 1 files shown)


tiagopereira:
Hi @DanRyanIrish, thanks for the comments.  I will put the change to add pix, but I think we should not put  all dimensions (the current formulation was was in fact already there). Using the first dimension might lead people to think that the data are arrays with shape `[n_repeats, n_raster, n_y, n_wave]`, while in fact they are a list of arrays with shape `[n_raster, n_y, n_wave]`. From the current metadata. I could also not find an obvious way to differentiate between sit-and-stare and multiple raster repeats. But I added a new line with the number of repeats.

DanRyanIrish:
Hi @tiagopereira.  Yes, that is a good point.  A single data array at the most basic level only ever stores data from a single raster scan and so is 3D.  But I still think there is a good argument for showing the 4 dimensions in the ```repr```, even though that's not what was originally there.

The data for each spectral window can be comprised of multiple raster scans that are stored as an ```IRISSpectrogramCubeSequence```.  This class has been specifically designed to appear to the user as if it were 4D.  So in the case of an OBS with 3 raster scans a user could do the following:
```python
>>> spectral_window_mgiik = my_irisspectrograph.data["mgiik"] # Extract mgiik data for convenience
>>> spectral_window_mgiik.dimensions # Get dimensions of spectral window data.
(<Quantity 3 pix>, <Quantity 8 pix>, <Quantity 548 pix>, <Quantity 333 pix>)
>>> spectral_window_mgiik.world_axis_physical_coords # Get the physical types of the axes.
('sequence axis', 'lon, 'lat', 'wavelength')
>>> mgiik_roi = spectral_window_mgiik[0:2, :, 200:300, 150:200] # Slice a region of interest out of the data.
>>> mgiik_roi.dimensions # Check dimensions of region of interest
(<Quantity 2 pix>, <Quantity 8 pix>, <Quantity 100 pix>, <Quantity 50 pix>)
```
So in practice, the user interacts with the data as if it were a 4D.  So I think it is more helpful to give the user all the dimensions in the ```repr``` as they wont really care whether the underlying data arrays are 3D, because they will use the object as though it is 4D.  The users are free to pull out the individual data arrays and hack them up if they want, but that no longer represents the structure of the ```IRISSpectrograph```.  So it's OK if that's inconsistent with the ```repr```.  What do you think?

As for the distinction between rasters, and sit-and-stare, I guess the metadata giving the number of unique raster positions works just fine.

irispy/spectrograph.py
```diff
@@ -54,18 +54,34 @@ def __init__(self, data, meta=None):
...
+                    end.strftime("%H:%M:%S.%f")))
+            else:
+                result.append("{0} -- {1}".format(start, end))
+        obs_period, instance_period = result
+        return ("<iris.IRISSpectrograph instance\nOBS ID: {obsid}\n"
+               "OBS Description: {obsdesc}\n"
+               "OBS period: {obs_period}\n"
+               "Instance period: {inst_period}\n"
+               "Number unique raster positions: {nraster}\n"
+               "Number of repeats: {repeats}\n"
```
DanRyanIrish: I think there's a case for going with something like ```OBS Number unique raster positions``` and ```OBS Number of repeats``` to make it clear these metadata refer to the the OBS as a whole, rather than the data in the ```IRISSpectrograph```.  If the data are sliced, or if only some of the files from the OBS were read in, these numbers will disagree with the shape of the data as the are taken from metadata and could be confusing.

tiagopereira:
I see. Makes sense if you can index the object with all axes. Changed to reflect latest suggestions.

irispy/spectrograph.py
```diff
@@ -73,11 +73,11 @@ def __repr__(self):
...
         return ("<iris.IRISSpectrograph instance\nOBS ID: {obsid}\n"
                "OBS Description: {obsdesc}\n"
-               "OBS period: {obs_period}\n"
+               "OBS Period: {obs_period}\n"
                "Instance period: {inst_period}\n"
-               "Number unique raster positions: {nraster}\n"
-               "Number of repeats: {repeats}\n"
-               "Spectral windows : dimensions [raster axis, slit axis, spectral axis]:"
+               "OBS Number unique raster positions: {nraster}\n"
+               "OBS Number of repeats: {repeats}\n"
```
DanRyanIrish: This value is now actually being taken from the data, rather than metadata describing the OBS as a whole.  So I think it should be ```"Instance Number of repeats"```.  In fact I think it's better to remove it altogether as the number of repeats in each instance is given by dimensions below.

DanRyanIrish:
Hi @tiagopereira.  I think this is looking really good now.  I left a final comment about the ```"OBS Number of repeats"``` line.  But other than that I'm happy to approve this PR once the online tests pass :)

DanRyanIrish:
Thanks @tiagopereira for improving the ```__repr__```!  The improvements are now in master!

Issue #72, opened 2018-04-26
New SJICubeSequence Class


DanRyanIrish:
The development of an ```SJICubeSequence``` class would provide a few important benefits.  Firstly, it would enable multiple SJI files to be read into a single object, say if a OBS was split over two files due an eclipse, or if a user wanted to combine data from different campaigns using the same OBS number.  Secondly, it would enable an ```SJICube``` to be broken out into a sequence of single images, each with its own WCS.  This would enable fine-scale pointing corrections to be made on a frame by frame basis, but still give users an object that they could use as a single cube.

Proposed Implementation
* ```SJICubeSequence``` would be subclassed from ```NDCubeSequence```
* As SJI data is always 3D (unlike spectrograph data), the ```__getitem__``` method would be over-ridden.  If the dimensionality of the sequence was 3D, i.e. all sub-cubes were 2D, normal slicing would be implemented.  If the dimensionality is 4D, i.e. sub-cubes are 3-D, the ```index_as_cube``` slicing property would be called.  This could also be determined by whether there was a common axis set.
* A new section of the ```read_iris_sji_level2_fits``` would be added to handle multiple files and return a ```SJICubeSequence```.

PR #83, opened 2018-05-07
Creation of IRISMapCubeSequence class

BaptistePellorceAstro:
Here the creation of the IRISMapCubeSequence class. See #72 for more informations.

### DIFF
--- a/irispy/new_sji.py
+++ b/irispy/new_sji.py
@@ -226,7 +226,7 @@ class IRISMapCubeSequence(NDCubeSequence):
     """
     def __init__(self, data_list, meta=None, common_axis=0):
         # Check that all SJI data are coming from the same OBS ID.
-        if len(np.unique([cube.meta["OBSID"] for cube in data_list])) != 1:
+        if np.any([cube.meta["OBSID"] != data_list[0].meta["OBSID"] for cube in data_list]):
             raise ValueError("Constituent IRISMapCube objects must have same "
                              "value of 'OBSID' in its meta.")
         # Initialize Sequence.
@@ -353,4 +353,7 @@ def read_iris_sji_level2_fits(filenames, memmap=False):
                                          extra_coords=extra_coords, scaled=scaled))
         hdulist.close()
 
-    return IRISMapCubeSequence(list_of_cubes, meta=meta, common_axis=0)
+    if len(filenames) == 1:
+        return list_of_cubes[0]
+    else:
+        return IRISMapCubeSequence(list_of_cubes, meta=meta, common_axis=0)
--- a/irispy/new_sji.py
+++ b/irispy/new_sji.py
@@ -274,7 +274,7 @@ Axis Types:\t\t {axis_types}
 
     def __getitem__(self, item):
         return utils.sequence.slice_sequence(self, item)
-        
+
     @property
     def cube_dimensions(self):
         return self.cube_like_dimensions
--- a/irispy/new_sji.py
+++ b/irispy/new_sji.py
@@ -12,6 +12,7 @@ from astropy.wcs import WCS
 from sunpy.time import parse_time
 from ndcube import NDCube
 from ndcube.utils.cube import convert_extra_coords_dict_to_input_format
+from ndcube import utils
 from ndcube.ndcube_sequence import NDCubeSequence
 
 from irispy import iris_tools
@@ -271,12 +272,14 @@ Axis Types:\t\t {axis_types}
            seq_shape=self.dimensions,
            axis_types=self.world_axis_physical_types)
 
-    @property
+
     def __getitem__(self, item):
-        return self.index_as_cube
+        return utils.sequence.slice_sequence(self, item)
+        #return self.index_as_cube
+        #return utils.sequence._index_sequence_as_cube(self, item)
 
     @property
-    def dimensions2(self):
+    def cube_dimensions(self):
         return self.cube_like_dimensions
 
     @property
--- a/irispy/new_sji.py
+++ b/irispy/new_sji.py
@@ -272,12 +272,9 @@ Axis Types:\t\t {axis_types}
            seq_shape=self.dimensions,
            axis_types=self.world_axis_physical_types)
 
-
     def __getitem__(self, item):
         return utils.sequence.slice_sequence(self, item)
-        #return self.index_as_cube
-        #return utils.sequence._index_sequence_as_cube(self, item)
-
+        
     @property
     def cube_dimensions(self):
         return self.cube_like_dimensions
--- a/irispy/new_sji.py
+++ b/irispy/new_sji.py
@@ -273,7 +273,7 @@ Axis Types:\t\t {axis_types}
            inst_start=instance_start,
            inst_end=instance_end,
            seq_shape=self.dimensions,
-           axis_types=self.world_axis_physical_coords)
+           axis_types=self.world_axis_physical_types)
 
     def __getitem__(self, item):
         return self.index_as_cube[item]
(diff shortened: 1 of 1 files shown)


irispy/new_sji.py
```diff
@@ -204,14 +205,82 @@ def apply_exposure_time_correction(self, undo=False, force=False):
...
+    meta: `dict` or header object
+        Metadata associated with the sequence.
+
+    common_axis: `int`
+        The axis of the NDCubes corresponding to time.
+
+    """
+    def __init__(self, data_list, meta=None, common_axis=0):
+        # Check that all SJI data are coming from the same OBS ID.
+        if len(np.unique([cube.meta["OBSID"] for cube in data_list])) != 1:
```
DanRyanIrish: This could be rewritten as:
```python
if any([cube.meta["OBSID"] for cube in data_list] != data_list[0].meta["OBSID"]):
```
The list might have to be an array.  I'm not sure.  Check this out.

irispy/new_sji.py
```diff
@@ -222,62 +291,66 @@ def read_iris_sji_level2_fits(filename, memmap=False):
...
+                'ENDOBS': endobs,
+                'NBFRAMES': hdulist[0].data.shape[0],
+                'OBSID': hdulist[0].header.get('OBSID', None),
+                'OBS_DESC': hdulist[0].header.get('OBS_DESC', None)}
+        list_of_cubes.append(IRISMapCube(data_nan_masked, wcs, uncertainty=uncertainty,
+                                         unit=unit, meta=meta, mask=mask,
+                                         extra_coords=extra_coords, scaled=scaled))
+        hdulist.close()
+
+    return IRISMapCubeSequence(list_of_cubes, meta=meta, common_axis=0)
```
DanRyanIrish: I think an ```IRISMapCubeSequence``` should only be returned if there is more than one file input.  Otherwise, an ```IRISMapCube``` should be returned as before.

BaptistePellorceAstro: Ok, this is boring to use it with only one file (I am working with). I will modify it.

irispy/new_sji.py
BaptistePellorceAstro: With this line, I get an error that I solve by doing :
<if np.any([cube.meta["OBSID"] != data_list[0].meta["OBSID"] for cube in data_list]):>

irispy/new_sji.py
```diff
@@ -204,14 +205,82 @@ def apply_exposure_time_correction(self, undo=False, force=False):
...
+def read_iris_sji_level2_fits(filenames, memmap=False):
     """
     Read IRIS level 2 SJI FITS from an OBS into an IRISMapCube instance
 
     Parameters
     ----------
-    filename : `str`
-        File name to be read
+    filenames: `list` of `str` or `str`
+        Filename of filenames to be read.  They must all be associated with the same
```
DanRyanIrish: Typo: "of" should be "or"

BaptistePellorceAstro:
Hi, as you said, I have some problems :'

 - `dimensions` : The problem is that `cube_like_dimensions` is calling `dimensions` so I have a circular error. I succeed to solve it by changing the name of my function to `dimensions2` for example. Do you have an idea to solve this bug ?

- `__getitem__` : For every tests I did, I have already this error `__getitem__() missing 1 required positional argument: 'item'`.

irispy/new_sji.py
```diff
@@ -272,12 +272,9 @@ def __repr__(self):
            seq_shape=self.dimensions,
            axis_types=self.world_axis_physical_types)
 
-
     def __getitem__(self, item):
         return utils.sequence.slice_sequence(self, item)
```
BaptistePellorceAstro: I use `utils.sequence.slice_sequence(self, item)` rather than `index_as_cube` because the last function is calling a lot of function that are not relevant (creating as many error as functions) and return the same result. What do you think about it ?

irispy/new_sji.py
```diff
@@ -274,7 +274,7 @@ def __repr__(self):
 
     def __getitem__(self, item):
         return utils.sequence.slice_sequence(self, item)
```
BaptistePellorceAstro: I use `utils.sequence.slice_sequence(self, item)` rather than `index_as_cube` because the last function is calling a lot of function that are not relevant (creating as many error as functions) and return the same result. What do you think about it ?

DanRyanIrish: I think you should call ```index_as_cube```.  The extra checks it does make it more rigorous.  We just need to get it working here.  And I think simply calling ```utils.sequence.slice_sequence``` will not always be correct.  The way you should call ```index_as_cube``` it is:
```python
return self.index_as_cube[item]
```
Have you tried this?  If so, what errors are you getting?

DanRyanIrish:
Hi @BaptistePellorceAstro.  I just noticed how long ago you sent your first comment since I last looked at this.  Apologies for the delay in replying.  I've left a couple comments to get you going again.  I'll also look into the ```cube_like_dimensions``` issue you asked about and get back to you.

DanRyanIrish:
@BaptistePellorceAstro, I think the solution to the ```dimensions``` issue will require a simple change to ```NDCubeSequence``` in the ndcube package:
* Create a new hidden method, ```NDCubeSequence._dimensions```, that is just a copy of the current ```dimensions``` method.
* Rewrite ```NDCubeSequence.dimensions``` so it simply calls the hidden version ```_dimensions```.
* In ```NDCubeSequence.cube_like_dimensions``` replace calls to the ```self.dimensions``` method with calls to the hidden version ```self._dimensions```.
* Then, back in IRISpy, you should be able to rewrite ```dimensions``` with a call to ```cube_like_dimensions``` as there will no longer be circular calling.

Does this make sense?

irispy/new_sji.py
BaptistePellorceAstro: Now it works !

irispy/new_sji.py
```diff
@@ -270,13 +273,13 @@ def __repr__(self):
...
     def __getitem__(self, item):
-        return utils.sequence.slice_sequence(self, item)
+        return self.index_as_cube[item]
 
     @property
-    def cube_dimensions(self):
+    def dimensions(self):
         return self.cube_like_dimensions
 
     @property
```
DanRyanIrish: Sorry, I think I told you to call this property ```world_axis_physical_coords```.  It was a typo.  It should have the same name as the ```NDCubeSequence``` method which is ```world_axis_physical_types```.  My mistake.  THis means it also needs changing in line 276 well.

irispy/new_sji.py
```diff
@@ -101,11 +103,14 @@ def __repr__(self):
         #Conversion of the end date of OBS
         endobs = self.meta.get("ENDOBS", None)
         endobs = endobs.isoformat() if endobs else None
-        #Conversion of the instance start of OBS
-        instance_start = self.extra_coords["TIME"]["value"][0]
+        #Conversion of the instance start and end of OBS
```
DanRyanIrish: Just so you know, the convention of comments is to use a space between the # and the comment.

DanRyanIrish:
Hi @BaptistePellorceAstro.  I think this PR is almost ready to merge! :) I just asked for a method name change that was originally my fault for giving you the wrong name.  See above comment.

irispy/new_sji.py

irispy/new_sji.py
BaptistePellorceAstro: Of course, I just forget to add a space.

BaptistePellorceAstro:
If everything is OK for you, I am ready to merge.

DanRyanIrish:
Another merged PR!  You're flying through them!

PR #84, opened 2018-05-07
Add code to convert and manipulate the IRIS OBSID

tiagopereira:
It would be nice to convert an IRIS OBSID number to a human readable format. And, ideally, finding out the modified OBSID if one changes parameters (e.g. exposure time).

This request adds, at the moment, some basic functionality to convert from an OBSID to a human readable printout. This is using the csv tables from LMSAL. Right now it only uses the strings directly from the csv, but it shouldn't be much more effort to convert to proper values when appropriate (e.g. "Deep x 8" > 8 sec).


DanRyanIrish:
Hi @tiagopereira.  This is a great idea.  I'll give the PR a review when I get a chance.  Also, don't forget we have an IRISpy meeting today.  It would be great for you and/or @BaptistePellorceAstro to attend as I'm hoping they will tell us about the dusk mask issue.

DanRyanIrish:
Hi @tiagopereira.  I had a quick look at this PR.  Could you explain what information the Obsid class is providing to the user, and how that should be accessed by the user?  Could you also explain a little bit about what the various tables are?  One first glance, the PR looks a bit complicated for what I think it's trying to do.  But I'm not confident I fully understand the subtleties of what it's doing either.

tiagopereira:
Hi @DanRyanIrish, the previous version had minimal docstrings, but this is hopefully better now. The idea is to deconstruct an IRIS OBSID into human readable values. I expect that most of the time users will just run it as a function and see the `__repr__`:

```
>>> obsid.Obsid(3675508564)
IRIS OBS ID 3675508564
----------------------
Description:            Large dense 96-step raster 31.35x120 96s
SJI filters:                                    C II   Mg II w s
SJI field of view:                                       120x120
Exposure time:                                               8.0 s
Binning:                               Spatial x 1, Spectral x 1
FUV binning:                         FUV spectrally rebinned x 4
SJI cadence:                                 SJI cadence default
Compression:                                Lossless compression
Linelist:                                       Flare linelist 1
```

But you can also get the parameters as attributes, e.g.:
```
>>> data = obsid.Obsid(3675508564)
>>> data.exptime
<Quantity 8. s>
>>> data.raster_fov
'31.35x175'
```

About the tables. These are the IRIS observing tables that the Table Tool uses  and available from http://www.lmsal.com/iris_science/planningconf  (see also [ITN 31](https://www.lmsal.com/iris_science/doc?cmd=dcur&proj_num=IS0205&file_type=pdf)). The format is not straightforward, hence the code seeming complicated. If you think there is a better way to do it, please let me know.

DanRyanIrish:
Hi @tiagopereira.  I think this is going to be a very useful piece of functionality!  I'm thinking about whether there is a simply way for users to access this.  Is there anything preventing the data being entered to the class as a dictionary?  That way the ```read``` method would not be needed.  Users could use the ```get``` method for dictionaries or access the data via ```Obsid.data["Description"]```.

In fact, the whole class could be inherited from ```dict``` and the ```__init__``` and ```__repr__``` could be altered as you have done here.  But that would mean that users could simply access the data by doing ```Obsid["Description"]```.

This PR also brings up another important, larger scale question.  That is how should we handle data files in IRISpy.  We should decide on a policy that we can carry forward consistently in the future.

tiagopereira:
Do you mean a dictionary with all the possible OBS IDs? If so, that would be millions of combinations. If not, then I don't see how you can get rid of the read function. If you mean just how it's accessed, should be ok with either dictionary or as attributes (although I prefer the latter as they require entering fewer characters). 

We could have the data from the CSV files saved in the class, but I generally think it is a bad idea to hardcode data inside code. Plus, if new tables are issued this code should be fairly easy to adapt. I wasn't sure what the policy was on data files, so I just put them under data since they fit the description in README.rst. If we are ever going to add data files in IRISpy, then I think these are very good candidates since they are so small.

DanRyanIrish:
You're right!  A dictionary of millions of combinations would be insane!  I didn't mean that.  I meant once the data for a single OBD was entered into the object, it could be as a dictionary.  The read function could then be called by, or within, the ```__init__``` function and enter the data into a  ``dict``` rather than attributes.  As understand, the ```read``` and/or ```__init__``` take the OBSID, figures out the relevant CSV tables, pulls the data in, processes what needs to be processed, then stores it in way the user can easily access and provides a useful ```__repr__```.  Am I understanding how this workflow happens?

Regarding attributes and dictionaries, I initially favoured attributes as well for the same reason.  Although I have altered my view after my experience with  ```sunpy.Map```.  There, everything is an attribute and the code becomes a beast to maintain.  It also means that if the tables change, there may be unused attributes that have to be removed, or new ones to add to avoid breaking.  Although, the dictionary interface does require a couple extra key strokes (and if using IPython tab complete it is only 2), the future maintenance work is reduced because you only have to alter the data read in code.  In fact if the read in algorithm is general enough, it may require no changes!  A dictionary also gives you a ```.keys()``` attribute which shows you all the data names.  Because by definition, this can't include any other methods or attributes it makes it easier to separate the data from other attributes when inspecting the object.  It also enables you to easily check whether the object contains a certain key and return a default value if not.  In addition, the ```.keys()``` method allows you to easily iterate over all keys.  Having said all this, attributes still have an important place, but I feel it's usually neater to pack up the data in an object like a dictionary or table.

Let me know if you think this workflow can/should be applied to this case or if the OBS data is too complicated in some way for this.

Regarding the CSV files, I think throwing them into the PR was a good first step to get the discussion going and to show show this OBSID class works.  Even if we don't leave them there.

tiagopereira:
Your understanding of how the code works is correct. 

It will be trivial to change from attributes to dictionary. If you look in the code, you'll see that the attributes are already dynamically set, so v36 tables will not have the same fields as v38 tables, and different tables have different "default string" for parameters, hence multiple keys having the same value in `self.field_keys`. This is also why some parameters do not appear in the `__repr__`.

irispy/obsid.py
```diff
@@ -0,0 +1,152 @@
...
+        """
+        Converts an 'exptime' string (used in IRIS tables and OBS_DESC)
+        to a Quantity instance in seconds.
+        """
+        if exptime == "Exposure 1s":
+            return 1. * u.s
+        else:
+            return float(exptime.split(' x ')[1]) * u.s
+
+    def read_obsid(self, obsid):
```
DanRyanIrish: This method should be hidden (```_read_obsid```) since, as I understand, this function is not supposed to be used directly by users, but rather called by the ```__init__```.  Also making it hidden means we give the user no guarantee that its API wont change in the future.  So if we want to rewrite it for whatever reason, we don't have to worry about deprecation warnings etc.  Easier for maintenance.

irispy/obsid.py
```diff
@@ -0,0 +1,152 @@
+# -*- coding: utf-8 -*-
+
+import numpy as np
+import pandas as pd
+from astropy import units as u
+from pkg_resources import resource_filename
+
+
+class Obsid(dict):
```
DanRyanIrish: Do you think the class should be renamed ```OBSID```?  If not, it should probably be at least ```ObsId``` or ```ObsID``` since classes should be camel-case.

irispy/obsid.py
```diff
@@ -0,0 +1,152 @@
...
+    'Flare linelist 1'
+    >>> data['raster_fov']
+    '31.35x120'
+    >>> data['raster_step']
+    0.33
+    """
+
+    def __init__(self, obsid):
+        self.obsid = obsid
+        self.field_keys = {'Large linelist': 'linelist',
```
DanRyanIrish: Now this object is dict-like, and the keys can be got by doing ```Obsid.keys()```, I don't think there's any need to store these as an attribute.  Why not just define it as a local variable ```field_keys``` inside ```read_obsid```?

irispy/tests/test_obsid.py
```diff
@@ -0,0 +1,36 @@
...
+TEST_DATA['sji_cadence'] = ['SJI cadence 0.5x faster', 'SJI cadence default',
+                            'SJI cadence 10s']
+TEST_DATA['linelist'] = ['Flare linelist 1', 'Full readout', 'Small linelist']
+
+
+@pytest.mark.parametrize("attr_name, test_input, expected_output",
+       [(name, obs, output[i]) for name, output in TEST_DATA.items()
+        for i, obs in enumerate(OBSID)])
+def test_attribute(attr_name, test_input, expected_output):
+    assert getattr(Obsid(test_input), attr_name) == expected_output
```
DanRyanIrish: Shouldn't this now be
```python
test_output = Obsid(test_input)
assert test_output[attr_name] == expected_output
```

DanRyanIrish:
Hi @tiagopereira.  I've added some minor comments to the PR.  But overall this is looking great!  I think the biggest blocker to merging this is now deciding how to deal with the data files.

DanRyanIrish:
Also, I noticed that this PR includes changes to spectrograph.py.  Is this intentional?  It looks like it might just be that you need to pull the latest changes from the main repo master branch into your PR branch as I think these changes were from your last PR?

tiagopereira:
Had a brief chat about this with @Cadair, who seems to be in favour of including the data files in the repository since they are so small.

Hopefully the issue with the changes in `spectrograph.py` is also solved.

DanRyanIrish:
OK.  Well I can get behind that.  One question beforehand though, are these files likely to ever be updated?  Or would any updates involve new tables be given a new version number and so would be a new file?  Like going from v38 to v36?  If this is the case, I don't see a big issue with versioning and so makes it much easier to just include the files.

tiagopereira:
New tables will be given in new files (two files per table version), so I think it is extremely unlikely they will every be updated. In the meanwhile, I also got some other files from Ryan Timmons for the calibration OBSIDs (series v42), but I need to look in more detail on how to include that. Much of the information in those new files is not of use for this purpose, and maybe it is better to add that as data inside the code. But probably better to close this PR and bring the calibration stuff in a new PR.

DanRyanIrish:
OK.  Then I think you're right that it's fine to just include these files in the repo.

I also agree that we should close out this PR and the issue of series v42 can be treated separately.

Are you happy for this PR to merged as is?

tiagopereira:
Looks good to me, I'm happy with a merge!

DanRyanIrish:
lgtm! 👍

DanRyanIrish:
Done!  Thanks so much for this contribution @tiagopereira.  I will definitely be using this class :)

DanRyanIrish:
I've been using this class this week, @tiagopereira.  So useful! :D

PR #87, opened 2018-05-22
Add Plot Method to IRISMapCubeSequence which Calls NDCubeSequence.plot_as_cube

BaptistePellorceAstro:
Here I overwrite `plot()` to redirect this command to `plot_as_cube()`.

This PR must be merged after PR[#123](https://github.com/sunpy/ndcube/pull/123) in ndcube, a bug will occurs if not.

### DIFF
--- a/irispy/new_sji.py
+++ b/irispy/new_sji.py
@@ -295,7 +295,7 @@ Axis Types:\t\t {axis_types}
         Based on the cube-like dimensionality of the sequence and value of plot_axis_indices
         kwarg, a Line/Image Plot/Animation is produced.
 
-         Parameters
+        Parameters
         ----------
         axes: `astropy.visualization.wcsaxes.core.WCSAxes` or ??? or None.
             The axes to plot onto. If None the current axes will be used.
--- a/irispy/new_sji.py
+++ b/irispy/new_sji.py
@@ -303,12 +303,12 @@ Axis Types:\t\t {axis_types}
         plot_axis_indices: `int` or iterable of one or two `int`.
             If two axis indices are given, the sequence is visualized as an image or
             2D animation, assuming the sequence has at least 2 cube-like dimensions.
-            The cube-like dimension indicated by the 0th index is displayed on the
-            x-axis while the cube-like dimension indicated by the 1st index is
-            displayed on the y-axis. If only one axis index is given (either as an int
-            or a list of one int), then a 1D line animation is produced with the indicated
-            cube-like dimension on the x-axis and other cube-like dimensions represented
-            by animations sliders.
+            The cube-like dimension indicated by the 0th element of plot_axis indices is
+            displayed on the x-axis while the cube-like dimension indicated by the 1st
+            element of plot_axis_indices is displayed on the y-axis. If only one axis
+            index is given (either as an int or a list of one int), then a 1D line
+            animation is produced with the indicated cube-like dimension on the x-axis
+            and other cube-like dimensions represented by animations sliders.
             Default=[-1, -2]. If sequence only has one cube-like dimension,
             plot_axis_indices is ignored and a static 1D line plot is produced.
 
--- a/irispy/new_sji.py
+++ b/irispy/new_sji.py
@@ -286,6 +286,12 @@ Axis Types:\t\t {axis_types}
     def world_axis_physical_types(self):
         return self.cube_like_world_axis_physical_types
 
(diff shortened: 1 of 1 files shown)


irispy/new_sji.py
```diff
@@ -286,6 +286,12 @@ def dimensions(self):
     def world_axis_physical_types(self):
         return self.cube_like_world_axis_physical_types
 
+    def plot(self, axes=None, plot_axis_indices=None, axes_coordinates=None,
+             axes_units=None, data_unit=None, **kwargs):
```
DanRyanIrish: This method needs a docstring so it's clear exactly how it works.  Without one, it will be confusing as it behaves differently to ```NDCubeSequence.plot```.  You can model the docstring of the ```NDCubeSequence.plot_as_cube``` docstring.  But you can simplify it as ```IRISMapCubeSequence``` has a fixed number of dimensions.

BaptistePellorceAstro:
I added a docstring but I think I have nothing to remove because the user may want to use different plots (1D, 2D or 3D). Tell me if you want I remove somethings.

irispy/new_sji.py
```diff
@@ -288,6 +288,77 @@ def world_axis_physical_types(self):
...
     def plot(self, axes=None, plot_axis_indices=None, axes_coordinates=None,
              axes_units=None, data_unit=None, **kwargs):
+        """
+        Visualizes data in the IRISMapCubeSequence with the sequence axis folded
+        into the common axis.
+
+        Based on the cube-like dimensionality of the sequence and value of plot_axis_indices
+        kwarg, a Line/Image Plot/Animation is produced.
+
+         Parameters
```
DanRyanIrish: Unwanted extra space before ```Parameters```.

irispy/new_sji.py
```diff
@@ -288,6 +288,77 @@ def world_axis_physical_types(self):
...
+        ----------
+        axes: `astropy.visualization.wcsaxes.core.WCSAxes` or ??? or None.
+            The axes to plot onto. If None the current axes will be used.
+
+        plot_axis_indices: `int` or iterable of one or two `int`.
+            If two axis indices are given, the sequence is visualized as an image or
+            2D animation, assuming the sequence has at least 2 cube-like dimensions.
+            The cube-like dimension indicated by the 0th index is displayed on the
+            x-axis while the cube-like dimension indicated by the 1st index is
```
DanRyanIrish: The x-axis index is the first element of ```plot_axis_indices```.  So ```plot_axis_indices=[0, 1]``` will have the 0th dimensions of the cube of the x-axis of the plot.  But ```plot_axis_indices=[1, 0]``` should have the axes inverted.

DanRyanIrish:
Looks pretty good.  Just made a couple minor comments.

irispy/new_sji.py
BaptistePellorceAstro: I think the text is correct because the 0th and 1st index are here the index of `plot_axis_indices`.

irispy/new_sji.py
BaptistePellorceAstro: Removed

irispy/new_sji.py
DanRyanIrish: Ah OK.  You're right.  But it is confusing.  Could you replace ```The cube-like dimension indicated by the 0th index is displayed...``` with something like ```The cube-like dimension indicated by the 0th index of plot_axis_indices is displayed...```?

DanRyanIrish: Or better yet, ```0th element of plot_axis_indices``` instead of ```0th index of plot_axis_indices```.

DanRyanIrish:
Barring that one request about the docstring, I think this is ready to go!

Oh, have you updated this PR with the latest version of master recently?  Not sure if there have been any PRs merged since you opened this PR.

DanRyanIrish:
Yes, you need do merge master since I just merged your other PR :)

DanRyanIrish:
Yes, I think this is looking good!  Now it just needs a couple of tests.  You can base them off the equivalent tests for ```IRISSpectrogram``` and ```IRISSpectrogramSequence```:
https://github.com/sunpy/irispy/blob/master/irispy/tests/test_spectrograph.py#L175
https://github.com/sunpy/irispy/blob/master/irispy/tests/test_spectrograph.py#L205

BaptistePellorceAstro:
You gave me the tests for `apply_exposure_time_correction` but we are the the `plot()` PR. Do you mean to create the tests for all methods inside `IRISMapCubeSequence` ?

PR #88, opened 2018-05-22
Add apply exposure time correction method to IRISMapCubeSequence

BaptistePellorceAstro:
Creating an exposure time correction method to `IRISMapCubeSequence`.

Some tests must be done before merging.

### DIFF
--- a/irispy/new_sji.py
+++ b/irispy/new_sji.py
@@ -286,6 +286,42 @@ Axis Types:\t\t {axis_types}
     def world_axis_physical_types(self):
         return self.cube_like_world_axis_physical_types
 
+def apply_exposure_time_correction(self, undo=False, force=False):
+        """
+        Applies or undoes exposure time correction to data and uncertainty and adjusts unit.
+
+        Correction is only applied (undone) if the object's unit doesn't (does)
+        already include inverse time.  This can be overridden so that correction
+        is applied (undone) regardless of unit by setting force=True.
+
+        Parameters
+        ----------
+        undo: `bool`
+            If False, exposure time correction is applied.
+            If True, exposure time correction is removed.
+            Default=False
+
+        force: `bool`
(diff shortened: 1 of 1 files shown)


irispy/new_sji.py
```diff
@@ -286,6 +286,42 @@ def dimensions(self):
...
+        result: `None` or `IRISMapCube`
+            If copy=False, the original IRISMapCube is modified with the exposure
+            time correction applied (undone).
+            If copy=True, a new IRISMapCube is returned with the correction
+            applied (undone).
+
+        """
+        new_seq = []
+        for cube in self.data:
+            new_seq.append(cube.apply_exposure_time_correction(undo=undo, force=force))
```
DanRyanIrish: This can be written as a list comprehension which is faster for long lists and the same speed for short lists: 
```python
corrected_data = [cube.apply_exposure_time_correction(undo=undo, force=force) for cube in self.data]
```

irispy/new_sji.py
```diff
@@ -286,6 +286,42 @@ def dimensions(self):
     def world_axis_physical_types(self):
         return self.cube_like_world_axis_physical_types
 
+    def apply_exposure_time_correction(self, undo=False, force=False):
```
DanRyanIrish: The API should mimic that of ```IRISSpectrogramCubeSequence.apply_exposure_time_correction```.  So there should also be a ```copy=False``` kwarg.  Then in the method, the ```self.data``` should be overwritten is ```copy=False``` and a new object returned if ```copy=True```.  See https://github.com/sunpy/irispy/blob/master/irispy/spectrograph.py#L174

DanRyanIrish:
Looks great!  Just added one minor comment.

DanRyanIrish:
👍 Another merged PR! :)

PR #89, opened 2018-05-24
Documentation on IRISpy

BaptistePellorceAstro:
Here the documentation about `IRISMapCube` and `IRISMapCubeSequence`.


BaptistePellorceAstro:
I saw, when I wrote that, the IRISMapCubeSequence is built like a list of IRISMapCube. So if you load a file at 1330A and one at 1400A, you will first see the full 1330A the the 1400A and so the time sequence is broken between the file. Do you prefer the current version or an other version where the time is conserved and where the frames can be : 
t0 => the first frame of 1330A
t1 => the first frame of 1400A

BaptistePellorceAstro:
For now, what do you think about that @DanRyanIrish ?

docs/IRISMapCubeSequence.rst
```diff
@@ -0,0 +1,103 @@
+===================
+IRISMapCubeSequence
+===================
+
+This class represents Slit Jaw Images from IRIS described by a single World Coordinate
+System (WCS). This class is a list of IRISMapCube objects, one per time step, from the
```
DanRyanIrish: This is not correct.  Each sub-cube is 2 or 3D described by a single WCS.  That can be a cube from a single file, or each exposure, for example if the user has used ```explode_along_axis```.

docs/IRISMapCubeSequence.rst
```diff
@@ -0,0 +1,103 @@
...
+same observation.
+
+Initialization
+--------------
+
+To initialize an IRISMapCubeSequence, we will need to open some level 2 fits file.
+You can find these files in the IRIS_ website by selecting a file either by clicking
+on the map or by setting some information in the left panel. Now that we get some level 2
+fits file, we can load it into the reading method for fits files. This method will read
+all the fits file and will take every information that it needs to create an
```
DanRyanIrish: ```every``` should be ```all the```

docs/IRISMapCube.rst
```diff
@@ -0,0 +1,94 @@
...
+System (WCS).
+
+Initialization
+--------------
+
+To initialize an IRISMapCube, we will need to open a level 2 fits file. You can find one
+of these files in the IRIS_ website by selecting a file either by clicking on the map or by
+setting some information in the left panel. Now that we get a level 2 fits file, we can
+load it into the reading method for fits files. This method will read all the fits file and
+will take every information that it needs to create an IRISMapCube object.
```
DanRyanIrish: ```every``` should be ```all the```

docs/IRISMapCube.rst
```diff
@@ -0,0 +1,94 @@
...
+
+Let assume that we will call our fits file ``my_fits_file`` and IRISMapCube object as
```
DanRyanIrish: Let's

docs/IRISMapCube.rst
```diff
@@ -0,0 +1,94 @@
...
+    Obs. End:			 2018-04-27T01:39:47.122000
+    Instance Start:		 2018-04-26T23:07:22.880000
+    Instance End:		 2018-04-27T01:36:40.490000
+    Total Frames in Obs.:	 49
+    IRIS Obs. id:		 3690015104
+    IRIS Obs. Description:	 Very large sit-and-stare 0.3x175 1s  C II   Si IV   Mg II h/k Deep x
+    Cube dimensions:		 [  49. 1095. 1018.] pix
+    Axis Types:			 ('time', 'custom:pos.helioprojective.lat', 'custom:pos.helioprojective.lon')
+
+Or we can print the data of our IRISMapCube object in an image or animation by doing that: ::
```
DanRyanIrish: ```print``` should be ```plot```

docs/IRISMapCube.rst
```diff
@@ -0,0 +1,94 @@
...
+
+    >>> my_cube.plot()
+
+We can upgrade the visualization by using kwargs to set value limits. Here, ``vmin`` is
+the minimum value and every lower value will be considered as the ``vmin`` value.
+This is the same for ``vmax`` that corresponds to the maximum value: ::
```
DanRyanIrish: Suggest replace with ```We can customize the visualization by using standard matplotlib kwargs relevant to the type of visualization produces by the plot method.  For example, for a 2D image/animation, we can use ```vmin``` and ```vmax``` to set the floor and ceiling of the color map like so: ::```

docs/IRISMapCube.rst
```diff
@@ -0,0 +1,94 @@
...
+
+    >>> my_cube.plot(vmin=0, vmax=300)
+
+Now that we have created our IRISMapCube object, we can use one of the followed methods
+to correct the data.
```
DanRyanIrish: ```correct``` should be ```manipulate```.

docs/IRISMapCube.rst
```diff
@@ -0,0 +1,94 @@
...
+kwarg by doing: ::
+
+    >>> from irispy import read_iris_sji_level2_fits
+    >>> my_cube = read_iris_sji_level2_fits(my_fits_file, memmap=True)
+
+So now, ``my_cube`` is an IRISMapCube object and we can access to a lot of information like:
+
+- ``my_cube.data`` : In this attribute, we can find the data array.
+- ``my_cube.wcs`` : This attribute contains the WCS.
+- ``my_cube.mask`` : This attribute can be used to mask some data (eg. the dust particle).
```
DanRyanIrish: There should be a paragraph discussing the uncertainty attribute and how the FITS reader function calculates it.

docs/IRISMapCubeSequence.rst
```diff
@@ -0,0 +1,103 @@
...
+IRISMapCubeSequence object.
+
+Let assume that we will build an IRISMapCubeSequence with two fits files that we will
```
DanRyanIrish: Let's

docs/IRISMapCubeSequence.rst
```diff
@@ -0,0 +1,103 @@
...
+call ``my_fits_file_0`` and ``my_fits_file_1``. We can create an IRISMapCubeSequence
+with as much fits files as we want but from the same observation ! Next, we will call
```
DanRyanIrish: ```much``` should be ```many```

Remove ``` !``` and replace with ```.```

docs/IRISMapCubeSequence.rst
```diff
@@ -0,0 +1,103 @@
...
+
+As an IRISMapCubeSequence object is a list of IRISMapCube objects, we can do the same things
+that we have seen previously in the IRISMapCube documentation. To access to the information
+stored in a IRISMapCube object, we can select the first IRISMapCube object by writting
+``my_sequence[0]`` or the Nth object with ``my_sequence[N-1]``. As they are IRISMapCube
+objects, we can access to their information by doing:
+
+- ``my_sequence[0].data`` : The data attribute of the first IRISMapCube object.
+- ``my_sequence[1].wcs`` : The WCS attribute of the second IRISMapCube object.
+- ``my_sequence[2].mask`` : The mask attribute of the third IRISMapCube object.
```
DanRyanIrish: And the uncertainty attribute?

docs/IRISMapCubeSequence.rst
```diff
@@ -0,0 +1,103 @@
...
+    Axis Types:		 ('time', 'custom:pos.helioprojective.lat', 'custom:pos.helioprojective.lon')
+
+Or we can print the data of our IRISMapCubeSequence object in an image or animation
+by doing that: ::
+
+    >>> my_sequence.plot()
+
+We can upgrade the visualization by using kwargs to set value limits. Here, ``vmin`` is
+the minimum value and every lower value will be considered as the ``vmin`` value.
+This is the same for ``vmax`` that corresponds to the maximum value. ::
```
DanRyanIrish: See comment regarding equivalent paragraph in ```IRISMapCube``` page.

docs/IRISMapCubeSequence.rst
```diff
@@ -0,0 +1,103 @@
...
+
+    >>> my_sequence.plot(vmin=0, vmax=300)
+
+Now that we have created our IRISMapCubeSequence object, we can use one of the followed methods
+to correct the data.
```
DanRyanIrish: ```correct``` should be ```manipulate```

DanRyanIrish:
Hi @BaptistePellorceAstro.  Here are some comments on your documentation.  In addition to those, I think there should be a subsection on the indexing, either via the standard slicing, or the ```crop_by_coords``` method.  You can draw a lot from what's already in the ```ndcube``` documentation

docs/IRISMapCubeSequence.rst
BaptistePellorceAstro: Modified

docs/IRISMapCube.rst
BaptistePellorceAstro: Modified

docs/IRISMapCube.rst
BaptistePellorceAstro: Modified

docs/IRISMapCube.rst
BaptistePellorceAstro: Modified

docs/IRISMapCube.rst
BaptistePellorceAstro: Replaced

docs/IRISMapCube.rst
BaptistePellorceAstro: Modified

docs/IRISMapCubeSequence.rst
BaptistePellorceAstro: Modified

docs/IRISMapCubeSequence.rst
BaptistePellorceAstro: Modified

docs/IRISMapCubeSequence.rst
BaptistePellorceAstro: Modified

docs/IRISMapCubeSequence.rst
BaptistePellorceAstro: Modified

docs/IRISMapCube.rst
BaptistePellorceAstro: Added

docs/IRISMapCubeSequence.rst
BaptistePellorceAstro: Added

docs/IRISMapCubeSequence.rst
BaptistePellorceAstro: Modified

DanRyanIrish:
I think that's a fine way to go.  But I think the docs should be broken into sections, e.g. ```Indexing```, ```Visualization```, even if they simply refer to the ndcube docs.  People shouldn't have to reread the whole thing to find what they need.  Sections will help the docs act as a tutorial for first time readers AND a reference for those who want to double check something.

DanRyanIrish:
A suggested structure for the ```IRISMapCube``` section would be:
* Initialization
* Class Structure (including where data, uncertainties, etc. can be found in object as well as extra coordinates like time)
* Dimensions
* Cropping & Indexing
  * Array-like Indexing
  * Cropping by Real World Coordinates
* Data Manipulation
  * Exposure Time Correction
  * Dust Mask

Many of these sections require a sentence and a link to the relevant ndcube doc page.  The others probably will require minor changes to paragraphs you've already written.

BaptistePellorceAstro:
Hi, I changed my mind. This should be easier to read and understand if we not link NDCube but copy it inside the documentation ?

docs/IRISMapCube.rst
```diff
@@ -0,0 +1,155 @@
...
+    >>> my_cube.apply_exposure_time_correction(undo=True, force=True)
+
+Dust particle mask
+^^^^^^^^^^^^^^^^^^
+
+If we take time to look inside the data, we can see that some spots are not relevant with
+the data we want to study. These spots are dust particles that can come from space or a
+not perfect construction of the satellite. In IRISMapCube, we can use a method that will
+modify the mask of our object with the dust particles positions. We can use this method
+by doing: ::
```
DanRyanIrish: Suggested rewrite: 

...inside the data, we see that some pixels are obscured by dust, and so do not reflect the emission from the Sun at that location.  The apply_dust_mask method of IRISMapCube can be used to add the locations of the dust pixels to the mask so that we can easily exclude them fro our analysis. 

>>> my_cube.apply_dust_mask()

docs/IRISMapCube.rst
```diff
@@ -0,0 +1,155 @@
...
+If the result is ``True``, the dust particles positions are applied in our ``my_cube.mask``.
+If the result is ``False``, the dust particles positions are not applied.
+
+Visualization
+-------------
+
+There is two different ways to visualize our IRISMapCube object. The first one is to use
+the representation of our object and the second one is to plot the data of our object.
+
+Representation
```
DanRyanIrish: I don't think this should be its own section, but should be included as part of the ```Initialization``` section.  It's not really visualization.  It's inspection.

docs/IRISMapCubeSequence.rst
```diff
@@ -0,0 +1,187 @@
...
+objects, we can access to their information by doing:
+
+- ``my_sequence[0].data`` : The data attribute of the first IRISMapCube object.
+- ``my_sequence[1].wcs`` : The WCS attribute of the second IRISMapCube object.
+- ``my_sequence[2].mask`` : The mask attribute of the third IRISMapCube object.
+- ``my_sequence[3].uncertainty`` : The uncertainty attribute of the fourth IRISMapCube object.
+
+Dimensions
+----------
+
```
DanRyanIrish: Since we overwrite the standard behaviour of ```NDCubeSequence.dimensions```, I think we should explicitly include a description and example of ```IRISMapCubeSequence.dimensions``` and ```IRISMapCubeSequence.world_axis_physical_types```

docs/IRISMapCubeSequence.rst
```diff
@@ -0,0 +1,187 @@
...
+Visualization
+-------------
+
+There is two different ways to visualize our IRISMapCubeSequence object. The first one
+is to use the representation of our object and the second one is to plot the data of
+our object.
+
+Representation
+^^^^^^^^^^^^^^
```
DanRyanIrish: Same comment as for the ```IRISMapCube``` version of this section.

DanRyanIrish:
Hi @BaptistePellorceAstro.  I agree with @tiagopereira that links to ndcube docs is much easier to maintain.  The one exception to this, as I've outlined in the line-by-line comments, is ```IRISMapCubeSequence.dimensions``` and ```IRISMapCubeSequence.world_axis_physical_types```.  Since we've overwritten the default ndcube behaviour, we should explictly explain how these work for ```IRISMapCubeSequence``` in the irispy docs.

docs/IRISMapCube.rst
BaptistePellorceAstro: Yes this is better, my sentences was horrible there !

docs/IRISMapCubeSequence.rst
BaptistePellorceAstro: Done. Effectively, this NDCube documentation don't describe our results.

docs/IRISMapCube.rst
BaptistePellorceAstro: Moved

BaptistePellorceAstro:
I have create sub-sections to explain more what are inside the classes, I think this this better like that but this is making the documentation longer.

tiagopereira:
I just fixed the language a little, also on `IRISMapCubeSequence.rst`. I noticed that there was a lot of duplication in `IRISMapCubeSequence.rst`, so I removed much of the text and linked with `IRISMapCube.rst` when relevant.

These two new files still have to be added to `index.rst`.

BaptistePellorceAstro:
Ok nice, I think this should be better than repeat everything.

docs/IRISMapCube.rst
```diff
@@ -4,72 +4,212 @@
...
+Our IRISMapCube object, here called ``my_cube``, is now created and inspected, we can look into its structure.
+
+Data
+^^^^
+
+The ``data`` attribute provides direct access to the data arrays. These values have a unit
+(see :ref:`Unit`), an uncertainty (see :ref:`Uncertainty`), and can be scaled or not
+(see :ref:`Scaled_Value`). A quick representation of these data can be seen by entering: ::
+
+    >>>my_cube.data
```
DanRyanIrish: Space needed after ```>>>```

docs/IRISMapCube.rst
```diff
@@ -4,72 +4,212 @@
...
+
+This will show a summary start and and of the data array. It such printouts it is common to see many ``nan`` values. This is because IRIS SJI images are often padded, and these values are masked (see :ref:`Mask`).
```
DanRyanIrish: We should remove reference to ```nan``` values as PR #93 will undo the assignment of ```nan``` values.  Perhaps this sentence could be something like:
```Note that IRIS SJI images are often padded with non-physical values, e.g. -200.  These pixels are identified by the IRISMapCube :ref:`Mask`.```

File: docs/IRISMapCube.rst [hunk 20]
```diff
@@ -4,72 +4,212 @@
...
+Here, we can find that our object has 3 axes: a helioprojective longitude, a helioprojective latitude and time. The dimension order is from the FITS file (written in Fortran order), and will be reversed when handling the arrays in Python (which uses C order). Therefore, time is usually the first dimension in the ``data`` array.
+
+The ``IRISMapCube`` object is aware of the WCS matrix as read from the FITS file: ``PC`` values represent the rotation matrix, ``CRVAL`` values represent coordinate values at the pixel positions of ``CRPIX``, and ``CDELT`` represent the pixel size in coordinate units. ``NAXIS`` represents the maximum dimension of each each axis.
+
+.. _Uncertainty:
+
+Uncertainty
+^^^^^^^^^^^
+
+We can also found another array inside our object, stored in the ``uncertainty`` attribute.
```
DanRyanIrish: ```found``` should be ```find```

docs/IRISMapCube.rst
```diff
@@ -4,72 +4,212 @@
...
+The uncertainty is calculated as the square root of our object data plus squared reading
+noise in photon unit. We can inspect the array by entering: ::
```
DanRyanIrish: ```unit``` should be ```units```

docs/IRISMapCube.rst
```diff
@@ -4,72 +4,212 @@
...
+
+    >>> my_cube.uncertainty
+
+This will return a summary of the uncertainty values. The ``nan`` values are
+also coming from the mask (see :ref:`Mask`).
```
DanRyanIrish: Also, remove reference to ```nan``` here too.

docs/IRISMapCube.rst
```diff
@@ -0,0 +1,264 @@
...
+
+If unsure if the the dust particle mask is applied
+to our ``my_cube.mask``, we can check the ``dust_masked`` attribute of our object: ::
+
+    >>> my_cube.dust_masked
+
+If ``True``, the dust particle positions are added to our ``my_cube.mask``, if
+``False`` the dust particle positions are not added.
+
+Visualisation
```
DanRyanIrish: American spelling should be used in accordance with the SunPy style rules, i.e. ```Visualization```.

docs/IRISMapCube.rst
```diff
@@ -0,0 +1,264 @@
...
+===========
+IRISMapCube
+===========
+
+This class provides access to IRIS Slit-jaw image (SJI) files.
+
+This class inherits from NDCube_, so many methods of IRISMapCube are linked to their
+description in NDCube_ documentation.
+
+Initialisation
```
DanRyanIrish: American spelling should be used, ```Initialization```.

docs/IRISMapCubeSequence.rst
```diff
@@ -0,0 +1,212 @@
...
+===================
+
+This class provides access to IRIS Slit-jaw image (SJI) files as a list of
+IRISMapCube objects, where these sub-cubes are 2D or 3D described by their
+ own World Coordinate System (WCS).
+
+This class inherits from NDCubeSequence_, so many methods of IRISMapCubeSequence are linked
+to their description in NDCubeSequence_ documentation.
+
+Initialisation
```
DanRyanIrish: American spelling should be used, ```Initialization```.

docs/IRISMapCubeSequence.rst
```diff
@@ -0,0 +1,212 @@
...
+--------------
+
+The initialisation of an IRISMapCubeSequence is very similar to that of :ref:`IRISMapCube`,
```
DanRyanIrish: American spelling should be used, ```Initialization```.

docs/IRISMapCubeSequence.rst
```diff
@@ -0,0 +1,212 @@
...
+via the function ``read_iris_sji_level2_fits``. This function takes as argument a file name (string) or a list of filenames.
+
+Let us assume that we want to load an IRISMapCubeSequence object with two fits files called ``my_fits_file_0`` and ``my_fits_file_1``. We can create an IRISMapCubeSequence
+with as many fits files as necessary, but ideally with from the same observation (e.g. observations interrupted by eclipses). To load an IRISMapCubeSequence object as ``my_sequence``, we do: ::
+
+    >>> from irispy import read_iris_sji_level2_fits
+    >>> my_fits_files = [my_fits_file_0, my_fits_file_1]
+    >>> my_sequence = read_iris_sji_level2_fits(my_fits_files)
+
+As for IRISMapCube, if you don't have a lot of RAM memory or if you are loading a huge file,
```
DanRyanIrish: I think this should be ```As with IRISMapCube``` as ```As for``` implies we are switching the focus of the sentence back to ```IRISMapCube```, rather than just referencing it.

docs/IRISMapCubeSequence.rst
```diff
@@ -0,0 +1,212 @@
...
+         IRIS Obs. id:		 3690015104
+         IRIS Obs. Description:	 Very large sit-and-stare 0.3x175 1s  C II   Si IV   Mg II h/k Deep x
+         Cube dimensions:		 [  49. 1095. 1018.] pix
+         Axis Types:			 ('time', 'custom:pos.helioprojective.lat', 'custom:pos.helioprojective.lon')
+         ]
+
+
+The return value is a list with the representation of each cube inside. For an upcoming work, we
+will need to select only on cube. We can do that by indexing our sequence from 0 to N-1,
+where N is the number of cubes. To select the first cube do: ::
```
DanRyanIrish: This sounds a bit confusing since although the data is split into two cubes, we've designed the slicing to allow users to treat it as one contiguous cube since the sequence axis has no physical meaning (unlike for spectrograph).  I'll need to think a little more about how best to word this.

BaptistePellorceAstro: Yes, I've ask you before but maybe the message went to an outdated review. I was asking you if we should mix the cubes and if yes in which order ? If we use time order, the wavelength will be mixed, opposite if we use wavelength (by wavelength I mean 2 different cube like cubes I used).

docs/IRISMapCubeSequence.rst
```diff
@@ -74,58 +83,58 @@ inside the object. We can see this list by doing: ::
...
+         IRIS Obs. id:		     3690015104
          IRIS Obs. Description:	 Very large sit-and-stare 0.3x175 1s  C II   Si IV   Mg II h/k Deep x
          Cube dimensions:		 [  49. 1095. 1018.] pix
          Axis Types:			 ('time', 'custom:pos.helioprojective.lat', 'custom:pos.helioprojective.lon')
          ]
 
-
 The return value is a list with the representation of each cube inside. For an upcoming work, we
 will need to select only on cube. We can do that by indexing our sequence from 0 to N-1,
 where N is the number of cubes. To select the first cube do: ::
```
BaptistePellorceAstro: Yes, I've ask you before but maybe the message went to an outdated review. I was asking you if we should mix the cubes and if yes in which order ? If we use time order, the wavelength will be mixed, opposite if we use wavelength (by wavelength I mean 2 different cube like cubes I used).

(rewrite out of outdated)

docs/IRISMapCubeSequence.rst
BaptistePellorceAstro: Yes this is what I said, but of course, I think that keeping the passband cubes is better. Do you want to regroup in only one cube the data ?

DanRyanIrish: No the data must physical reside in separate cubes because they have different WCS transformations.  But the user should interact with the sequence like it's a cube.

Yes, we definitely want to avoid that.

BaptistePellorceAstro: If I summarize, you prefer creating a sequence with `explode_along_axis` than what I did, and then, keep only the same passband. So maybe we must add so code in `IRISMapCubeSequence` to use only one passband ?

DanRyanIrish: Not quite.  You can use different FITS files, so long as they are from the same passband and have the same OBS ID.  This can happen in a couple of ways, like if a single observing campaign is interrupted by an eclipse (i.e. the Earth get's between the Sun and IRIS), or if the same observing mode (OBD ID) is used for two observing campaigns, e.g. if a long term study is carried out.  But by far the more common is due to eclipses breaking up a single observing campaign into two FITS files.

```explode_along_axis``` is valid way of making a sequence out of one FITS file.

I'm not sure we need to raise an error in ```IRISMapCubeSequence``` at this stage.  Maybe as the object gets used we will find users who like to use it that way.  But I think there should be an error raised by ```read_iris_sji_level2_fits``` if there isn't already.

DanRyanIrish: Let's have the errors raised in the FITS reading function for now.

BaptistePellorceAstro:
Hi @DanRyanIrish, as I will use this documentation in my internship report, I want to know how you feel the new version before adding it to finish my report.

DanRyanIrish:
Hi @BaptistePellorceAstro.  I'm sorry I haven't had a proper chance to read through it while travelling since I made the last comments.  But if you've made the changes suggested by my last comments I think it must be pretty good :)

Issue #82, opened 2018-05-04
Make loading IRISSpectrograph more memory efficient


tiagopereira:
I'm finding that `IRISSpectrograph`, or `read_iris_spectrograph_level2_fits()` takes a very long time to load and has a very high memory usage. Loading a moderate size IRIS raster in my laptop, which only has 8 GB of RAM, leads to lots of swapping and the machine nearly crawls to a halt. 

Testing in different machines, I find the RAM that `read_iris_spectrograph_level2_fits()` uses is between 3-4 times the total size of the IRIS level 2 FITS file. A 606 MB file took about 1.5 GB of RAM, while a 2.6 GB file took 10GB. This is very inefficient, not to mention slow (even if the machine has enough RAM). 

Looking in the code, I suspect these lines are the most problematic:
https://github.com/sunpy/irispy/blob/bbd194205c1dd815230925ad676a286f572bb28c/irispy/spectrograph.py#L562-L564 

There should be a way to load it with `memmap`, and ideally it should be the default.

In addition, the uncertainty will perform many operations on all data points from the file:

https://github.com/sunpy/irispy/blob/bbd194205c1dd815230925ad676a286f572bb28c/irispy/spectrograph.py#L596-L598

This is a huge burden for something of marginal value. 

The loading of `IRISSpectrograph` should be as lightweight as possible. People may use it just to look at what's inside.

DanRyanIrish:
Hi @tiagopereira.  I've opened a PR irradicating the ```deepcopy``` you highlighted above in a hope of reducing the RAM requirements of the ```read_iris_spectrograph_level2_fits()```.  Could you give it a quick review and play it with to see if it improves the cases you tested at all?

Another thing we could do is have the uncertainty derivation be optional so the user can turn it off if they don't have a lot of RAM.  This will require some changes to other parts of the code base where it is assumed ```uncertainty``` will always exist.  But it could be done.  We could then also add a ```derive_uncertainty``` method which populates this attribute if the user later decides they want it, say after they have extracted a region of interest and RAM considerations are no longer so important.  I can open a separate PR for that change when I get a chance.

Does this sound like a good way forward for now?

tiagopereira:
Hi @DanRyanIrish, thanks for this. I don't have any time this week but will look at this next week. Please ping me again if I forget about this.

tiagopereira:
[...]

PR #93, opened 2018-06-01
Remove conversion of -200 to NaN in read_iris_spectrograph_level2_fits.

DanRyanIrish:
This PR reduces the RAM needed during reading spectrograph files without changing the final output, except that masked values remain -200 rather than replaced with NaNs.  This operation was highlighted as a significant RAM drain in issue #82.


DanRyanIrish:
Hi @tiagopereira.  Thanks for the analysis of this PR.  It's very insightful.  Sounds to me like this PR is an improvement but there is clearly more to be done beyond the scope of this PR.  Could you copy your analysis in Issue #82?  I'll merge this PR and then we can continue the conversation about ```uncertainty``` and ```memmap``` in Issue #82 and implement improvements in new PRs.

PR #94, opened 2018-06-05
Solving the sunpy.map.GenericMap bug in sji.py

BaptistePellorceAstro:
This PR solve the current bug in sji.py due to the use of `sunpy.map.GenericMap` which is no longer imported.


DanRyanIrish:
Thanks for catching this @BaptistePellorceAstro.

PR #97, opened 2018-06-14
Replacing test_sji.py with the test_new_sji.py file

BaptistePellorceAstro:
@DanRyanIrish, can you explain me how I can delete the `test_new_sji.py` file from the repository ?


DanRyanIrish:
This is just deleting old test.sji.py and renaming test_new_sji.py to test_sji.py, right?

BaptistePellorceAstro:
Yes but I don't know if there is something special to remove the old file.

DanRyanIrish:
It should be as simple as deleting the old test_sji.py and renaming tes_new_sji.py.  It looks like that's what you've done.

BaptistePellorceAstro:
I cut what were in new_sji and paste it in sji but the new_sji is still here and as a blank file. I don't know how commit the deleting of a file.

DanRyanIrish:
Oh.  Delete it like normal and then do
```console
git add test_new_sji.py
git commit -m 'Removed now defunct test_new_sji.py file.'
```
and push the change to you branch.  Unfortunately I just merged it so open another PR from the same branch.

PR #99, opened 2018-06-24
Propagate scaled attribute through IRISMapCube slicing

DanRyanIrish:
This PR alters the slicing of ```IRISMapCube``` so that the ```scaled``` attribute is propagated.


DanRyanIrish:
Hi @BaptistePellorceAstro.  Could you give this a quick review?  It's only a couple line change.

DanRyanIrish:
Thanks @BaptistePellorceAstro

Issue #78, opened 2018-05-02
Dealing with inconsistent WCS data


tiagopereira:
Perhaps I'm just unlucky, but the first time I try to use the ndcube spectrograph class I get this error:

```
>>> s = read_iris_spectrograph_level2_fits("iris_l2_20180210_033712_3690115104_raster_t000_r00000.fits")
(...)
SingularMatrixError: ERROR 3 in wcsset() at line 2218 of file cextern/wcslib/C/wcs.c:
Linear transformation matrix is singular.
ERROR 3 in linset() at line 679 of file cextern/wcslib/C/lin.c:
PCi_ja matrix is singular.
```
I traced the error and it seems like `WCS` from `ndcube.utils.wcs` does not like the WCS matrix from that file, which is:

```
PC1_1   =        1.00000000000 /
PC1_2   =        0.00000000000 /
PC2_1   =        0.00000000000 /
PC2_2   =       0.999938666821 /
PC3_1   =        0.00000000000 /
PC3_2   =      0.0112707177177 /
PC3_3   =       0.999938666821 /
PC2_3   =     -0.0112707177177 /
```

This does NOT seem singular to me. In any case, `read_iris_spectrograph_level2_fits` should not bail out. It should deal gracefully with the error and continue (even if without WCS).

DanRyanIrish:
Is the problem here:  ```PC2_3   =     -0.0112707177177 /```?  Should this be ```PC3_2```?

DanRyanIrish:
Ah, I thought it was SJI. Never mind then.

tiagopereira:
I found the problem here. It affects all sit and stare raster files, and happens because CDELT3=0. It seems that `astropy.wcs.WCS` does not like this and throws a singular matrix error. I think in these cases `CTYPE3` should be `Time` instead of `HPLN-TAN`, but we can't change IRIS's format, so must find a workaround instead. I don't know what is best. A simple fix would be to set `CDELT=1e-10` or something. We could also set `CTYPE3=TIME`.

DanRyanIrish:
Thanks @tiagopereira for figuring this out. That does sound like you had to go digging in the weeds!

I agree with what say about the difference between the ideal and practical solutions. I’m not sure what the best way forward is either. I’ll give it sone thought. But perhaps your idea of setting CDELT3 to 1e-10 will end up being the best though...

steinhh:
[...]

PR #106, opened 2018-09-26
Bug Fix for Reading Sit-and-stare WCS info

DanRyanIrish:
This PR addresses Issue #78 which reveals that the CDELT3 value of sit-and-stare files is 0.  This causes a SingularMatrixError when reading the WCS info into an astropy WCS object.  The solution is to change CDELT to 1e-10 in these cases.

In addition, this PR changes the output when reading single sit-and-stare files to ```IRISSpectrogramCube``` for each spectral window rather than an ```IRISSpectrogramCubeSequence``` since single sit-and-stare files are 3D, not 4D like for raster data.

### DIFF
--- a/irispy/spectrograph.py
+++ b/irispy/spectrograph.py
@@ -552,10 +552,18 @@ def read_iris_spectrograph_level2_fits(filenames, spectral_windows=None):
         ophaseix = hdulist[-2].data[:, hdulist[-2].header["OPHASEIX"]]
         exposure_times_fuv = hdulist[-2].data[:, hdulist[-2].header["EXPTIMEF"]] * u.s
         exposure_times_nuv = hdulist[-2].data[:, hdulist[-2].header["EXPTIMEN"]] * u.s
-        general_extra_coords = [("time", 0, times), ("raster position", 0, raster_positions),
-                                ("pztx", 0, pztx), ("pzty", 0, pzty),
-                                ("xcenix", 0, xcenix), ("ycenix", 0, ycenix),
-                                ("obs_vrix", 0, obs_vrix), ("ophaseix", 0, ophaseix)]
+        # If OBS is raster, include raster positions.  Otherwise don't.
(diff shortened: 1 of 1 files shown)


DanRyanIrish:
ping @tiagopereira.  This PR addresses issue #78.  Will you give it a quick code review?  Feel free to take a look as well @byrdie

irispy/spectrograph.py
```diff
@@ -611,10 +623,18 @@ def read_iris_spectrograph_level2_fits(filenames, spectral_windows=None):
...
                                     mask=data_mask))
         hdulist.close()
-    # Construct dictionary of IRISSpectrogramCubeSequences for spectral windows
-    data = dict([(window_name, IRISSpectrogramCubeSequence(data_dict[window_name],
-                                                           window_metas[window_name],
-                                                           common_axis=0))
+    # Construct dictionary of IRISSpectrogramCubeSequences for spectral windows.
+    # The one exception is when a single sit-and-stare file is being read in.
+    # In that case create a dictionary of IRISSpectrogramCubes
+    if top_meta["NRASTERP"] == len(filenames) == 1:
```
byrdie: Is it really best to treat these two cases differently? To me it makes the most sense to leave the sit-and-stare observations in an `IRISSpectrogramCubeSequences` but with only one element. That way, both cases can be treated the same by the user.

DanRyanIrish: Good point.  I can remove this change.

DanRyanIrish: I have reverted the commit that made this change.  So the WCS fix will still work but the output type will be the same whether the file is sit-and-stare or raster.

tiagopereira:
Hi @DanRyanIrish, sorry I haven't had time to look at this. Travelling and currently at a course for the rest of the week.

DanRyanIrish:
No problem @tiagopereira!

PR #109, opened 2019-05-24
Applied Template

nabobalis:
Latest version now with a run of the linters I like on top. 

No more travis but means circle and azure need setting up.


ez_setup.py
```diff
@@ -54,7 +54,9 @@ def _python_cmd(*args):
```
Cadair: this can go, why is this in the template :nauseated_face:

nabobalis: It wasn't deleted?

irispy/extern/__init__.py
```diff
@@ -0,0 +1,9 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
```
Cadair: Do we need to be writing an extern if we don't have anything to put it to it? is it a cookie option?

nabobalis: There is a c extern option but not a normal extern one.

nabobalis:
It seems that not all the changes I was able to copy over.

nabobalis:
TODO: SQUASH THE COMMITS, FIX CIRCLECI, Probably some more

nabobalis:
I think this is now good to go at least for a merge and fix issues after point of view.

PR #114, opened 2019-08-01
Return nothing when trying to read empty list of spectra

tiagopereira:
Currently `read_iris_spectrograph_level2_fits` is giving an `UnboundLocalError` when given an empty list as argument. This is because `spectral_windows_req` is not defined in that case:

```
In [50]: spp = read_iris_spectrograph_level2_fits([])
UnboundLocalError                         Traceback (most recent call last)
<ipython-input-50-81c39e132216> in <module>
----> 1 spp = read_iris_spectrograph_level2_fits([])

~/codes/irispy/irispy/spectrograph.py in read_iris_spectrograph_level2_fits(filenames, spectral_windows)
    628                                                            window_metas[window_name],
    629                                                            common_axis=0))
--> 630                  for window_name in spectral_windows_req])
    631     # Initialize an IRISSpectrograph object.
    632     return IRISSpectrograph(data, meta=top_meta)

UnboundLocalError: local variable 'spectral_windows_req' referenced before assignment
```

This pull request checks if the argument is empty, then it will return nothing. We can also raise an exception instead.


DanRyanIrish:
Hi @tiagopereira.  Thanks for pointing out this behaviour.  I suggest we raise an error rather than return nothing.  Then a simple ```try...except...``` statement around ```read_iris_spectrograph_level2_fits``` can be used to achieve the same behavior as returning nothing without potentially confusing users who aren't expecting that behaviour.  I'll leave a suggested change along these lines.

irispy/spectrograph.py
```diff
@@ -468,6 +468,8 @@ def read_iris_spectrograph_level2_fits(filenames, spectral_windows=None):
     """
     if type(filenames) is str:
         filenames = [filenames]
+    if not filenames:
```
DanRyanIrish: Suggest these two lines be replaced with something like:
```python
if isinstance(filenames, list):
    raise TypeError("filenames must be a string or list of strings giving the filenames to read.")
if len(filenames) == 0:
    raise ValueError("No filenames provided.")
```

tiagopereira: Ok. Replace them. I kept the "if not filenames" as this is the recommended by PEP8 for checking for empty lists.

tiagopereira: Ok, I replaced the lines according to your suggestion. But I kept the `if not filenames` form, as this is recommended by PEP for checking empty lists.

irispy/spectrograph.py
```diff
@@ -486,8 +486,11 @@ def read_iris_spectrograph_level2_fits(filenames, spectral_windows=None):
     """
     if type(filenames) is str:
         filenames = [filenames]
+    if isinstance(filenames, list):
+        raise TypeError("filenames must be a string or list of strings giving"
+                        " the filenames to read.")
```
DanRyanIrish: Hi @tiagopereira.  Won't this raise an error when ```filenames``` is a ```list``` even though ```filenames``` is allowed to be a ```list```?  Are you trying check that all elements are strings in the case that ```filenames``` is a list?

tiagopereira:
It was so long since there was any activity here that I accidentally deleted the source repository.

PR #116, opened 2019-08-02
Add options to read_iris_spectrograph which makes it not use much memory

Cadair:


irispy/spectrograph.py
```diff
@@ -447,7 +447,7 @@ def apply_exposure_time_correction(self, undo=False, force=False):
             mask=self.mask, missing_axis=self.missing_axis)
 
 
-def read_iris_spectrograph_level2_fits(filenames, spectral_windows=None):
+def read_iris_spectrograph_level2_fits(filenames, spectral_windows=None, uncertainty=True, memmap=True):
```
Cadair: memmap should probably default to False because of `do_not_scale_image_data`

Cadair:
The best solution to this function is to use Dask, because then the image scaling and all the uncertainty and mask calculations can be done in a delayed fashion, which means if you don't need to use them they wont do anything. I am going to be looking at implementing some dask support pretty close to `astropy.io.fits` but this is a good stop gap.

We need this for https://github.com/glue-viz/glue-solar or else it eats all RAM everywhere.

tiagopereira:
This PR would also be very helpful in running IRIS demos from Binder, where the RAM is typically limited to 1-2 GB. I am planning on running a Binder-based IRIS python tutorial at the next IRIS workshop (Nov 4), so if this is merged by then I could include IRISpy.

irispy/spectrograph.py
DanRyanIrish: Yes, I think it makes sense to keep the same default behaviour as fits.io.

Cadair:
@DanRyanIrish updated it to be false by default.

tiagopereira:
It would be nice if this could be merged as soon as possible. I'm giving a tutorial using IRIS with python at the next IRIS meeting in about 3 weeks, and I'll be running it from Binder (1 GB RAM limitation). If this is merged, then I can use IRISpy. Otherwise I'll just stick to reading the FITS files by hand.

Cadair:
@tiagopereira I have pulled in master here and will merge if the CI passes.

@DanRyanIrish Would it be worth getting a release out soon?

PR #120, opened 2019-08-31
Fix default version code bug

kakirastern:
This should fix the bug left from the piece of code used to default to version 4 from a previous PR. It turns out that the conditional used was missing a crucial part which renders local tests invalid.


DanRyanIrish:
Hi @kakirastern.  Thanks for spotting and fixing this bug!

PR #129, opened 2020-03-09
Reformatted package

nabobalis:
Probably more to do.


DanRyanIrish:
Wow!  This a huge amount of work!  Thanks @nabobalis.

The only comment I have is that we don't want the ```instr``` directory removed at this point.  We may be able to keep the IRIS data classes in a similar way that sunpy not only has ```GenericMap``` but also ```AIAMap```.  A final decision on that hasn't been made, but I think we shouldn't use this PR to force it.  The two of us should discuss the future of IRIS code, hopefully at the sunpy coordination meeting.

nabobalis:
In that case I can add them back.

DanRyanIrish:
Awesome!

nabobalis:
I've added it that all back and circle seems to be running now.

DanRyanIrish:
Great!  Thanks.

Issue #137, opened 2020-03-29
Remove IRIS-specific code except for spectrograph reader


DanRyanIrish:
All iris-specific code should be moved to ```sunkit-instruments``` pending consensus that that's where it should live.  In the meantime, the IRIS-code will remain available in sunraster on the IRISpy branch.

The one IRIS function that should be kept is the spectrograph reader function.  This needs to be edited to read the data into ```Raster```, ```RasterSequence```, and ```NDCollection``` objects.
This approach will lead to the loss of the ```IRISSpectrograph.spectral_windows``` property but the spectral window names can be easily obtained from ```NDCollection.keys()```.

PR #147, opened 2020-04-11
Remove IRIS code except spectrograph reader.

DanRyanIrish:
This PR removes IRIS-specific code except for the IRIS spectrograph reader function.  This was placed in a new module ```io/iris.py```.  This PR also edits the reader function to return the data in generic ```sunraster``` and ```ndcube``` objects.

Addresses Issue #137 

To Do:
- [x] Make IRIS reader output generic data types: ```SpectrogramCube```, ```RasterSequence```, ```NDCollection``` (instead of ```IRISpectrograph```).


kakirastern:
Looks like for the `ci/circleci: html-docs` we got stuck with the docstring bug from `ndcube`...

kakirastern:
It looks like one change I have made to the line immediately following L407 in file `data/data_classes.rst` by adding the following didn't get picked up: 
```
['slit step' None 'spectral']
```
I know this looks a bit awkward, but without it the test on `ci/circleci: html-docs` could not get passed, at least on my laptop.

kakirastern:
Have added this PR in case it is needed: https://github.com/DanRyanIrish/sunraster/pull/7

DanRyanIrish:
> It looks like one change I have made to the line immediately following L407 in file `data/data_classes.rst` by adding the following didn't get picked up:
> 
> ```
> ['slit step' None 'spectral']
> ```
> 
> I know this looks a bit awkward, but without it the test on `sunpy.sunraster` could not get passed, at least on my laptop.

I deliberately didn't include this addition because the print statement that caused it has been deleted from the code.  So it is not output in the doctests anymore...or at least it shouldn't!

PR #159, opened 2020-04-14
non-conda on rtd

nabobalis:


nabobalis:
@DanRyanIrish might fix RTD

PR #163, opened 2020-04-20
Backport 0.1: Bump min ndcube version to 1.3.2

DanRyanIrish:


changelog/162.bugfix.rst
```diff
@@ -0,0 +1 @@
+Bump min ndcube version to fix bug caused when OS is bot 64-bit.
```
nabobalis, line 1:
```suggestion
Bump min ndcube version to fix bug caused when OS is not 64-bit.
```

PR #171, opened 2020-07-08
SPICE metadata object

DanRyanIrish:
This PR introduces metadata objects that define a standard API for accessing commonly used solar metadata for slit spectrograph instruments.  It includes a specific implementation for SolO/SPICE.


DanRyanIrish:
@Cadair, you might be interested in discussing this as I think it has potential implications for sunpy.

sunraster/meta/meta.py
```diff
@@ -0,0 +1,52 @@
+__all__ = ["RemoteSensorMetaABC", "SlitSpectrographMetaABC"]
+
+class RemoteSensorMetaABC():
```
Cadair, line 3: you should make this an actual ABC and use abstract method on the properties.

sunraster/meta/meta.py
```diff
@@ -0,0 +1,52 @@
...
+    @property
+    def dsun(self):
+        pass
+    
+    @property
+    def observer_coordinate(self):
+        pass
+    
+    @property
+    def date(self):
```
Cadair, line 37: `date` is probably not a good key, something like`observation_date` or even better `date_start` `date_end` would map better to FITS and sanity.

sunraster/meta/meta.py
```diff
@@ -0,0 +1,52 @@
...
+    @property
+    def rsun_meters(self):
+        pass
+    
+    @property
+    def rsun_obs(self):
+        pass
+    
+    @property
+    def dsun(self):
```
Cadair, line 29: `distance_to_sun`?

sunraster/meta/meta.py
```diff
@@ -0,0 +1,52 @@
...
+    @property
+    def processing_level(self):
+        pass
+    
+    @property
+    def rsun_meters(self):
+        pass
+    
+    @property
+    def rsun_obs(self):
```
Cadair, line 25: this is just FITS, not sure what's better though.

sunraster/meta/spice.py
```diff
@@ -0,0 +1,106 @@
+
+from astropy.io import fits
+import astropy.units as u
+
+
+__all__ = ["SPICEMeta"]
+
+
+class SPICEMeta(fits.header.Header, SlitSpectrographMetaABC):
```
Cadair, line 9: I am not sure inherting from fits header is wise, it might be better long term to keep the header as an attribute rather than actually *this* object.

DanRyanIrish:
@Cadair, does this now provide the features you were looking for?

PR #174, opened 2020-11-20
Remove dev test envs

DanRyanIrish:
The Dev dependencies, in particular ndcube, is currently incompatible with the stable version currently used by sunraster 0.2.  Therefore these test envs are removed.


nabobalis:
How large of a refactor needs to occur for sunraster to support the next version of ndcube?

Cadair:
many

PR #179, opened 2021-01-14
Fixed arguments in SpectrogramCube to match NDCube

tiagopereira:
### Description

Fixes the interface of `SpectrogramCube` so that it works with recent versions of `ndcube`.

### DIFF
--- a/changelog/179.bugfix.rst
+++ b/changelog/179.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed arguments in SpectrogramCube to match NDCube.
--- a/sunraster/spectrogram.py
+++ b/sunraster/spectrogram.py
@@ -166,8 +166,8 @@ class SpectrogramCube(NDCube, SpectrogramABC):
     def __init__(self, data, wcs, extra_coords=None, unit=None, uncertainty=None, meta=None,
                  mask=None, instrument_axes=None, copy=False, **kwargs):
         # Initialize SpectrogramCube.
-        super().__init__(data, wcs, uncertainty=uncertainty, mask=mask, meta=meta, unit=unit,
-                         extra_coords=extra_coords, copy=copy, **kwargs)
+        super().__init__(data, wcs=wcs, uncertainty=uncertainty, mask=mask, meta=meta, 
+                         unit=unit, extra_coords=extra_coords, copy=copy, **kwargs)
 
         # Determine labels and location of each key real world coordinate.
         self_extra_coords = self.extra_coords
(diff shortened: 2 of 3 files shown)


DanRyanIrish:
Thanks so much for this PRn @tiagopereira.  Will review now.

DanRyanIrish (changes requested):
A similar change is also needed here: https://github.com/sunpy/sunraster/pull/179/files#diff-700a3462d608adc85ab8a5e9cdec91e374215a187101945d745d0f98ad207964L169

nabobalis:
I would suggest we need a unit test of some sort to make sure this doesn't happen again.

DanRyanIrish:
Agreed.  I'm looking into that.  @tiagopereira, could you also go through the `SpectrogramCube`s created at the top of the [test file](https://github.com/sunpy/sunraster/blob/master/sunraster/tests/test_spectrogram.py) and make they are all instantiated with kwargs?  Then could you also add a mask to [spectrogram_instrument_axes](https://github.com/sunpy/sunraster/blob/master/sunraster/tests/test_spectrogram.py#L88).  I think that should catch it.

tiagopereira (commented):
Done

DanRyanIrish:
Great!  Just realised this isn't quite enough to catch this error if it happens again.  We have a test for confirming the `instrument_axes` are [sliced properly](https://github.com/sunpy/sunraster/blob/master/sunraster/tests/test_spectrogram.py#L164), but not one to check the other attributes are.  Can you add one more test that slices `spectrogram_instrument_axes` and then calls [this ndcube function](https://github.com/sunpy/ndcube/blob/master/ndcube/tests/helpers.py#L91) that checks all the attributes are as expected?  It will require instantiating the expected cube.  But it shouldn't take much time to do.

Also, can you replace [this print statement](https://github.com/sunpy/sunraster/blob/master/sunraster/tests/test_spectrogram.py#L167)?

tiagopereira:
I won't have time to work more on this for at least three weeks, but feel free to add the changes yourself if you want.

DanRyanIrish:
@tiagopereira, I opened a PR to your branch that (I hope) fixes the outstanding issues on this PR.  I realise you don't have much time at the moment, but if you could quickly merge it on your end then hopefully I will be able to merge this PR and we can move ahead with another sunraster release.  Alternatively, I can close this PR and open another so I have merge rights.  Let me know which you prefer.

DanRyanIrish (approved):

DanRyanIrish:
Merged!  Thanks so much @tiagopereira for this PR.  I plan to release sunraster 0.2 in the coming weeks so this bugfix will be available.

PR #187, opened 2021-11-03
Update exposure_time properties to use exposure time from Meta object.

DanRyanIrish:
Hi @nabobalis.  This PR is the result of trying to answer your comment question [here](https://github.com/sunpy/sunraster/pull/184/commits/1bf85774a063e6696313c8a60eda383e619266d8#diff-700a3462d608adc85ab8a5e9cdec91e374215a187101945d745d0f98ad207964R436) regarding why you had to transpose the data.

The answer is that transposing did not fix the problem.  It just enabled it to give a wrong answer without erroring.  The real issue was that the method found the spectral axis when is should have been finding the axis to which the exposure time corresponded.  Fixing that bug then led me to update the code and tests to work when exposure time was stored in the new `Meta` instance.

### Description

Fixes #<Issue Number>

### DIFF
--- a/sunraster/tests/test_spectrogram.py
+++ b/sunraster/tests/test_spectrogram.py
@@ -7,6 +7,7 @@ from ndcube.tests.helpers import assert_cubes_equal
 
 import sunraster.spectrogram
 from sunraster import SpectrogramCube
+from sunraster.extern.meta import Meta
 
 # Define a sample wcs object
 H0 = {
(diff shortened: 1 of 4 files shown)


sunraster/tests/test_spectrogram.py
```diff
@@ -74,69 +75,71 @@
         (Time("2017-01-01") + TimeDelta(np.arange(TIME_DIM_LEN, TIME_DIM_LEN * 2), format="sec")),
     ),
 ]
+meta_exposure0 = Meta({"exposure time": EXPOSURE_TIME}, axes={"exposure time": 0},
+                      data_shape=SOURCE_DATA_DN.shape)
```
nabobalis, line 79: This is something we expect users to write in future?

nabobalis:
Thanks for fixing my mistakes!

PR #204, opened 2022-06-08
Fixes Issue 196 - read_spice_l2_fits function fails to load wide-slit SPICE FITS files

EdBehn:
## PR Description

This change allows the function read_spice_l2_fits to load wide-slit SPICE FITS files. 
Previously, this failed because the wavelength axis is not exclusively associated with wavelength. (It is mixed with HP Latitude and Longitude.) Now, alignment of said axis is not enforced. 

See: https://github.com/sunpy/sunraster/issues/196

### DIFF
--- a/sunraster/instr/spice.py
+++ b/sunraster/instr/spice.py
@@ -111,7 +111,7 @@ def read_spice_l2_fits(filenames, windows=None, memmap=True, read_dumbbells=Fals
             aligned_axes = tuple(range(len(first_sequence.dimensions)))
         else:
             aligned_axes = tuple(
-                i for i, phys_type in enumerate(first_sequence.array_axis_physical_types) if phys_type != ("em.wl",)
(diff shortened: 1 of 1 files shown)


nabobalis:
Thanks for the PR, it looks good to me, I will need to see about adding a test in some form to ensure that sunraster does not break this feature again.

changelog/204.bugfix.rst
```diff
@@ -0,0 +1 @@
+- Allow SPICE FITS reader to read handle wide-slit files. (`#204 <https://github.com/sunpy/sunraster/pull/204>`__)
```
nabobalis, line 1:
```suggestion
Allow SPICE FITS reader to read handle wide-slit files.
```
These get added when the changelog is created, so no need to worry about them for now!

nabobalis:
What would be the best way to shrink a level 2 file to use as a test file for the reader? The original issue links a 300MB file which is a bit too large to use. If I were to drop all but the first HDU, would that still suffice?

PR #254, opened 2024-01-17
correcting units in spice.py

grahamkerr:
The units of the SPICE level 2 were hardcoded in spice.py as u.adu, whereas these can be obtained from the fits header (L2 should be W/m^2/sr/nm. The u.adu was causing some issues when dealing with NDUncertainty.


nabobalis (approved):

alasdairwilson (approved):

nabobalis:
Thanks @grahamkerr, I will get a bugfix release out now.

PR #273, opened 2025-02-03
update ndcube

nabobalis:


sunraster/spectrogram.py
```diff
@@ -311,40 +311,13 @@ def __str__(self):
...
-        except KeyError:
-            pass
-        return result
+    @property
+    def instrument_axes(self):
+        """
+        The relationship between the array axes and the instrument,
+        i.e. repeat raster axis, slit position, position along slit, and spectral.
+        """
+        return self.meta["instrument_axes"]
```
DanRyanIrish, line 320:
```suggestion
        return self.meta.get("instrument_axes")
```

nabobalis:
Is the API that they must be NDMeta or can dicts be passed in? And if so, should those not be cast to NDMeta?

